PR #5, opened 2019-10-04
Initial support for Shippable.

mattclay:
Initial support for Shippable based on the existing https://github.com/ansible/ansible/ setup.


gavinfish:
@mattclay Should I wait for a stable version of Ansible?

mattclay:
I think this is ready to be merged. There will need to be some follow-up updates later:

1. When the next Ansible 2.9 RC is released we should switch to that release. The same for when 2.9.0 is released.
2. As more integration tests are added to the collection that use additional CI groups, those additional groups will need to be added to `shippable.yml` for those tests to run.

mattclay:
You can see that groups 1, 2, 5 and 7 are currently in `shippable.yml` on this PR:

https://github.com/ansible/ansible_collections_azure/blob/b27bfa3a51c83d37f9a1343d7c030a4372d175cf/shippable.yml#L13-L23

Those are the groups currently used by the tests in this repository. Here's one example:

https://github.com/ansible/ansible_collections_azure/blob/63dbb154ee95b38326061bcf67f404e3bd247135/test/integration/targets/azure_rm_deployment/aliases#L3

So when additional tests are added, if those tests use another group (there are currently 10 in the ansible/ansible repo for Azure tests), then those group(s) will need to be added to `shippable.yml` to run those tests. Otherwise they will be ignored by Shippable.

mattclay:
The version of Ansible used for the tests can be updated here:

https://github.com/ansible/ansible_collections_azure/blob/b27bfa3a51c83d37f9a1343d7c030a4372d175cf/tests/utils/shippable/shippable.sh#L76

thedoubl3j:
thanks for adding the comments in @mattclay

PR #9, opened 2019-11-05
Bump test ansible version to v2.9.0

gavinfish:


tests/utils/shippable/shippable.sh
```diff
@@ -73,7 +73,7 @@ set +ux
 . ~/ansible-venv/bin/activate
 set -ux
 
-pip install git+https://github.com/ansible/ansible.git@d0a4f0fdbe3f54dd029430f0c943df4ec9fb5877 --disable-pip-version-check
+pip install git+https://github.com/ansible/ansible.git@v2.9.0 --disable-pip-version-check
```
mattclay: No need to install from git now that the install is switching to a released version:

```suggestion
pip install ansible==2.9.0 --disable-pip-version-check
```

PR #11, opened 2019-11-13
Inventory Plugin

willtome:
Add inventory plugin from ansible/ansible to be distributed with collection


gavinfish:
@willtome Thanks! Several concerns on my side.

1. Any tests for this? How should an inventory plugin be tested in collection?
2. Basic documentation is also necessary to tell users how to use this instead of the origin one.

willtome:
@gavinfish currently there are not tests for this inventory plugin in Ansible core. Looking at how this is being done with AWS (https://github.com/ansible/ansible/tree/devel/test/integration/targets/inventory_aws_ec2) it should be a similar pattern to write for Azure assuming CI is set up the same. 

The plugin is documented in the code `DOCUMENTATION` section. This is used to generate the official documentation (https://docs.ansible.com/ansible/latest/plugins/inventory/azure_rm.html). I'm happy to add to the readme but do not want to duplicate the inherent documentation mechanism for plugins.

gavinfish:
@willtome Yes, similar tests need to be done.

I think a brief description to tell the difference of using origin azure inventory plugin and collection azure inventory plugin needs to be added in the readme.

gavinfish:
@willtome it seems the shippable job is stuck, please reset it before I can merge this PR.

willtome:
@gavinfish shippable looks like it has completed. I’m not sure what the other check is or how to change it.

Issue #13, opened 2019-11-14
remove deprecated content


thedoubl3j:
The `_azure.py` module should be removed as it is an empty file that was deprecated and removed in ansible version 2.8.

I would recommend not shipping anything that is deprecated in your collection, especially the one going up to Automation Hub. 

I will open a PR that removes the file shortly.

gavinfish:
Fixed in #14 .

PR #14, opened 2019-11-14
removing deprecated content and fixing galaxy.yml

thedoubl3j:
This PR removes the deprecated module `_azure.py` as well as update the galaxy.yml file with removing the license_file: `key`. Ran into an issue that appeared on automation hub that was not found in galaxy. It can be found here - https://github.com/ansible/galaxy/issues/2102

This will get the collection in a shippable state for automation hub as everything else looked good on our end. :)


thedoubl3j:
fixes #13

PR #19, opened 2019-12-03
Fix inventory plugin NAME

willtome:
fix `NAME` of plugin to support collection


willtome:
@gavinfish one bug I just caught was the name of the plugin that the auto plugin uses to source the correct file. This PR fixes it. Tests updated. still unsure how to complete last check.

PR #75, opened 2020-03-17
Use hasattr instead of dir

atombrella:
##### SUMMARY

I believe that this is better. I don't think that there should be cases where `value` would not be part of `item`.
https://stackoverflow.com/questions/17723569/whats-the-difference-between-hasattr-and-attribute-in-dir

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME

storageblob

##### ADDITIONAL INFORMATION


Fred-sun:
@haiyuazhang ready_for_review

PR #99, opened 2020-04-22
Fixing rule_type reference

Zype76:
Reference to ev instead of item is preventing the expected rule_type values from functioning

##### SUMMARY
When reviewing https://github.com/ansible/ansible/issues/68744 , I found that the "rule_type" if statement checks 'ev' instead of 'item'. This is preventing it from being parsed properly, and the _snake_to_camel function isn't triggered to filter the options into the desired format. 

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
The update will ensure that the expected rule_type values will function with the module.

##### ADDITIONAL INFORMATION
Testing: 
```
Before: 
"msg": "Error creating the Application Gateway instance: Azure Error: InvalidRequestFormat\nMessage: Cannot parse the request.\nException Details:\n\tError Code: InvalidJson\n\tMessage: Error converting value \"path_based_routing\" to type 'Microsoft.WindowsAzure.Networking.Nrp.Frontend.Contract.Csm.Public.ApplicationGatewayRequestRoutingRuleType'. Path 'properties.requestRoutingRules[0].properties.ruleType', line 1, position 1565."
}

After: 
TASK [Create instance of Application Gateway] ************************************************************************************************************************************************
task path: /home/zephyr/azure/aztest.yml:7

```


Fred-sun:
ready_for_review

PR #105, opened 2020-05-05
Make inventory_hostname configurable with hostvar_expressions

### LINKED ISSUE #104: Please allow me to generate the inventory_hostname
##### SUMMARY
I would like to define my own formula for generating the inventory_hostname. As Azure does not allow renaming VMs, I need to use a tag to get our names for the VM. I want to use our standardize tag name so that I can add vars in `host_vars/` as required.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm inventory plugin

##### ADDITIONAL INFORMATION
This is a continuation of https://github.com/ansible/ansible/issues/54065 which only requested the ability to configure the legacy behavior.

I want to define my own inventory_hostname as mentiond here:
https://github.com/ansible-collections/azure/blob/4467d13b459dc171857e3d7956a3f1fa6139d379/plugins/inventory/azure_rm.py#L366-L368

For reference, this is the section that currently defines the legacy/unique hostname definition:
https://github.com/ansible-collections/azure/blob/4467d13b459dc171857e3d7956a3f1fa6139d379/plugins/inventory/azure_rm.py#L504-L508

In both cases, this uses `vm_model['name']` but I explicitly do NOT want to use `vm_model['name']` because Azure doesn't let me rename VMs.

cognifloyd:
#### SUMMARY
Allow configuring inventory_hostname with a constructable/compose expression in hostvar_expressions.

Currently, you can add inventory_hostname to compose vars, but that is too late, as the hostname is already set. If you try, ansible with override it with a magic var to ensure that the var matches the hostname.

Resolves #104

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
azure_rm inventory plugin

##### ADDITIONAL INFORMATION
Example of how this could be used with tags:
```yaml
hostvar_expressions:
  inventory_hostname: tags.FQDN | default(default_inventory_hostname)
```

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -261,11 +261,13 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
 
         constructable_config_strict = boolean(self.get_option('fail_on_template_errors'))
         constructable_config_compose = self.get_option('hostvar_expressions')
+        constructable_config_inventory_hostname = constructable_config_compose.pop('inventory_hostname')
         constructable_config_groups = self.get_option('conditional_groups')
         constructable_config_keyed_groups = self.get_option('keyed_groups')
 
         for h in self._hosts:
-            inventory_hostname = self._get_hostname(h)
+            # FUTURE: track hostnames to warn if a hostname is repeated (can happen for legacy and for composed inventory_hostname)
+            inventory_hostname = self._get_hostname(h, compose=constructable_config_inventory_hostname, strict=constructable_config_strict)
             if self._filter_host(inventory_hostname, h.hostvars):
                 continue
             self.inventory.add_host(inventory_hostname)
@@ -298,9 +300,15 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
 
         return False
 
-    def _get_hostname(self, host):
-        # FUTURE: configurable hostname sources
-        return host.default_inventory_hostname
+    def _get_hostname(self, host, compose=None, strict=False):
+        if not compose:
+            return host.default_inventory_hostname
+        try:
+            return self._compose(compose, host.hostvars)
+        except Exception as e:
+            if strict:
+                raise AnsibleParserError("Error generating inventory_hostname with '{0}' (hostvar_expressions.inventory_hostname) for host {1}: {2}".format(compose, host.default_inventory_hostname, to_native(e))
+            return host.default_inventory_hostname
 
     def _process_queue_serial(self):
         try:
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -507,7 +507,8 @@ class AzureHost(object):
             ) if self._vmss else {},
             virtual_machine_size=self._vm_model['properties']['hardwareProfile']['vmSize'] if self._vm_model['properties'].get('hardwareProfile') else None,
             plan=self._vm_model['properties']['plan']['name'] if self._vm_model['properties'].get('plan') else None,
-            resource_group=parse_resource_id(self._vm_model['id']).get('resource_group').lower()
+            resource_group=parse_resource_id(self._vm_model['id']).get('resource_group').lower(),
+            default_inventory_hostname=self.default_inventory_hostname,
         )
 
         # set nic-related values from the primary NIC first
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -17,18 +17,6 @@ DOCUMENTATION = r'''
         - By default, sets C(ansible_host) to the first public IP address found (preferring the primary NIC). If no
           public IPs are found, the first private IP (also preferring the primary NIC). The default may be overridden
           via C(hostvar_expressions); see examples.
-        hostnames:
-          description:
-            - A list of Jinja2 expressions in order of precedence to compose inventory_hostname.
-            - Ignores expression if result is an empty string or None value.
-            - By default, inventory_hostname is generated to be globally unique based on the VM host name.
-              See C(plain_host_names) for more details on the default.
-            - An expression of 'default' will force using the default hostname generator if no previous
-              hostname expression resulted in a valid hostname.
-            - Use ``default_inventory_hostname`` to access the default hostname generator's value in
-              any of the Jinja2 expressions.
-          type: list
-          default: [default]
 '''
 
 EXAMPLES = '''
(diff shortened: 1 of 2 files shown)


cognifloyd:
I see some basic `azure_rm` tests in https://github.com/ansible-collections/azure/tree/dev/tests/integration/targets/inventory_azure, but that only tests the legacy_hostnames behavior. Do you need tests for this? If so, it is not clear how to add tests or how they fit with the existing tests.

Also, is shippable even running? I don't see this PR (or anything else) running at https://app.shippable.com/github/ansible-collections/azure/dashboard

mwinfie:
Can a maintainer review this? I think this would solve a similar issue we are having where we want to be able to pull in hosts by ip address instead of the vm name.

plugins/inventory/azure_rm.py
```diff
@@ -138,6 +138,9 @@
   # overrides the default ansible_host value with a custom Jinja2 expression, in this case, the first DNS hostname, or
   # if none are found, the first public IP address.
   ansible_host: (public_dns_hostnames + public_ipv4_addresses) | first
+  # overrides the default inventory_hostname value with a custom Jinja2 expression, in this case, a tag called vm_name.
+  # if the tag is not found, use the default_inventory_hostname (see plain_host_names).
+  inventory_hostname: tags.vm_name | default(default_inventory_hostname)
```
AlanCoding: I'm not saying that this is a good idea at all, but if you do this, you should at least put an item in the documentation for `hostvar_expressions` that explains this feature.

AlanCoding:
I know that `inventory_hostname` is a magic variable in playbooks, but I would, in no way, expect that setting that hostvar would change the name of the host. That's not have I would expect magic variables will work, and I know it doesn't jive with how the inventory plugin interface is written.

Several collections have some sort of option for doing this (modifying the name as it appears to Ansible, but not the connection name).

https://github.com/ansible-collections/amazon.aws/blob/fba558015ea763e025380d54c41b3fe30e4ffb42/plugins/inventory/aws_ec2.py#L41

https://github.com/ansible-collections/google.cloud/blob/8d231272a1523f9d137ba582483cb02eb27c733e/plugins/inventory/gcp_compute.py#L45

https://github.com/ansible-collections/vmware/blob/dd8add6d842ea595f34a461e0216a01501b8f874/plugins/inventory/vmware_vm_inventory.py#L68

You'll notice that in every one of these cases, there is a top-level option for the inventory plugin called `hostnames`. I would strongly suggest naming it the same yourself, not because it's the best possible name, but because it's the most standard.

I would also suggest mimicking the implementation of those other collections to the point of outright copying and pasting where possible. If you want to make it "nicer", you can see my most recent PR to VMWare https://github.com/ansible-collections/vmware/pull/253, which is for error handling, and will account for some differences with other examples.

cognifloyd:
Thanks for the review. I'll take a look at the interface for changing hostnames in the other inventory plugins.
 
This was the quick and dirty implementation, now let's see if we can get a better interface to provide this feature. Without it, I'm hard-coding my ansible inventory instead of using this plugin.

cognifloyd:
K. I copied from the vmware inventory with a few adjustments. And I rebased.

plugins/inventory/azure_rm.py
```diff
@@ -363,9 +383,30 @@ def _filter_host(self, inventory_hostname, hostvars):
...
-    def _get_hostname(self, host):
-        # FUTURE: configurable hostname sources
-        return host.default_inventory_hostname
+    def _get_hostname(self, host, hostnames=None, strict=False):
+        hostname = None
+        errors = []
+
+        for preference in hostnames:
+            if preference == 'default':
+                return host.default_inventory_hostname
```
AlanCoding, line 315: 👍  nice handling of existing behavior.

AlanCoding:
checked out your branch and tried running, error:

> name 'to_text' is not defined

applied

```diff
diff --git a/plugins/inventory/azure_rm.py b/plugins/inventory/azure_rm.py
index 899ba14..9e4050a 100644
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -198,7 +198,7 @@ from ansible.module_utils.six import iteritems
 from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common import AzureRMAuth
 from ansible.errors import AnsibleParserError, AnsibleError
 from ansible.module_utils.parsing.convert_bool import boolean
-from ansible.module_utils._text import to_native, to_bytes
+from ansible.module_utils._text import to_native, to_bytes, to_text
 from itertools import chain
 from msrest import ServiceClient, Serializer, Deserializer
 from msrestazure import AzureConfiguration
```

With that, I applied something similar to your example, confirmed use of custom name from tags for 1 host, rest of hosts still using default. So with that patch 👍  from me.

cognifloyd:
Rebased on dev.

AlanCoding:
I have ~any~ no objections, this seems like it's adopting a relatively common pattern for inventory plugins.

cognifloyd:
@AlanCoding I think you meant "I have no objections" instead of "I have any objections". Right?

AlanCoding:
yes, I have no objections.

Fred-sun:
@cognifloyd  Recently, I didn’t merge this PR because I was busy with other things. Can you help resolve the conflicting files? After the settlement, we will merge this PR as much as possible. Thank you!

cognifloyd:
rebased

plugins/inventory/azure_rm.py
```diff
@@ -17,6 +17,18 @@
...
+            - A list of Jinja2 expressions in order of precedence to compose inventory_hostname.
+            - Ignores expression if result is an empty string or None value.
+            - By default, inventory_hostname is generated to be globally unique based on the VM host name.
+              See C(plain_host_names) for more details on the default.
+            - An expression of 'default' will force using the default hostname generator if no previous
+              hostname expression resulted in a valid hostname.
+            - Use ``default_inventory_hostname`` to access the default hostname generator's value in
+              any of the Jinja2 expressions.
+          type: list
+          default: [default]
```
Fred-sun: Lines 20~31 should be added to plugins/doc_fragments/azure_rm.py, thank you!

cognifloyd: Done

laingsc:
I see this was merged into version 1.20.  Is this actually using hostvar_expressions (inventory_hostname), or is it using hostnames: as a list var?  Neither seems to work using latest 1.20:

```
---
include_vmss_resource_groups:
- '*'
hostvar_expressions:
  inventory_hostname: tags.Hostname | default(default_inventory_hostname)
default_host_filters: ["powerstate != \"running\""]
keyed_groups:
- prefix: tag
  key: tags

Or this:
---
include_vmss_resource_groups:
- '*'
hostnames:
  - tags.Hostname
default_host_filters: ["powerstate != \"running\""]
keyed_groups:
- prefix: tag
  key: tags
```

AlanCoding:
Example given is

```yaml
hostnames:
  - tags.vm_name
  - default  # special var that uses the default hashed name
```

which is similar to your 2nd attempt but not quite the same. The "tags" key should be in hostvars, print with `ansible-inventory`, and the sub-key needs to exist inside of that for it to work.

Issue #103, opened 2020-05-04
Inventory plugin does not work with Ansible 2.10 (ansible-base)


AlanCoding:
##### SUMMARY
The inventory plugin works with Ansible 2.9, but does not work with current `devel` branch, which becomes Ansible 2.10 at some point.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/inventory

##### ANSIBLE VERSION
```paste below
ansible-inventory 2.10.0.dev0
  config file = /etc/ansible/ansible.cfg
  configured module search path = [u'/var/lib/awx/.ansible/plugins/modules', u'/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python2.7/site-packages/ansible
  executable location = /usr/bin/ansible-inventory
  python version = 2.7.5 (default, Sep 26 2019, 13:23:47) [GCC 4.8.5 20150623 (Red Hat 4.8.5-39)]
```

##### CONFIGURATION
defaults

##### OS / ENVIRONMENT
mac / control machine

##### STEPS TO REPRODUCE
Call `ansible-inventory` and point it to your inventory file location.

##### EXPECTED RESULTS
works same in Ansible 2.9 and Ansible 2.10.

##### ACTUAL RESULTS

```
[WARNING]:  * Failed to parse /tmp/awx_135_3xm2x23k/azure_rm.yml with auto
plugin: unknown doc_fragment(s) in file /var/lib/awx/vendor/inventory_collectio
ns/ansible_collections/azure/azcollection/plugins/inventory/azure_rm.py: azure
  File "/usr/lib/python2.7/site-packages/ansible/inventory/manager.py", line 287, in parse_source
    plugin.parse(self._inventory, self._loader, source, cache=cache)
  File "/usr/lib/python2.7/site-packages/ansible/plugins/inventory/auto.py", line 50, in parse
    plugin = inventory_loader.get(plugin_name)
  File "/usr/lib/python2.7/site-packages/ansible/plugins/loader.py", line 553, in get
    self._load_config_defs(name, self._module_cache[path], path)
  File "/usr/lib/python2.7/site-packages/ansible/plugins/loader.py", line 290, in _load_config_defs
    add_fragments(dstring, path, fragment_loader=fragment_loader)
  File "/usr/lib/python2.7/site-packages/ansible/utils/plugin_docs.py", line 116, in add_fragments
[...]

PR #115, opened 2020-05-14
Migrate missing doc_fragments that went missing to this collection

AlanCoding:
##### SUMMARY
See issue https://github.com/ansible-collections/azure/issues/103

This pulls in:

https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/plugins/doc_fragments/azure.py
https://github.com/ansible/ansible/blob/stable-2.9/lib/ansible/plugins/doc_fragments/azure_tags.py

##### ISSUE TYPE
- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME
doc_fragments/azure.py
doc_fragments/azure_tags.py

##### ADDITIONAL INFORMATION
I tried to do due diligence asking here:

https://github.com/ansible/ansible/pull/67684#issuecomment-628095773

About whether or not the current code in this collection will be allowed to work. There was some confusion on that point, but they have clearly answered that nothing is planned that will support the non-FQCN doc fragment. I see no way to get this fixed in any reasonable timeframe other than the collection, itself, making this change.

Example of testing this:

```
ANSIBLE_COLLECTIONS_PATHS=../../.. ansible-doc azure.azcollection.azure_rm_image
```

This renders correctly (using Ansible `devel` branch), and makes use of both doc fragments. I have also tested the inventory plugin with `ansible-inventory`.

Ping @gundalow, because I know he has tried to do a sweep to check for items that went missing in the great collection migration. This isn't the first time this kind of situation has happened. Doc fragments were not around when several collections started out, so there's a clear story of how it got here.

if you consult the Ansible core file (the current file) `.github//BOTMETA.yml`, then you will find:

```
  lib/ansible/plugins/doc_fragments/azure.py:
    migrated_to: azure.azcollection
  lib/ansible/plugins/doc_fragments/azure_tags.py:
    migrated_to: azure.azcollection
```

So clearly _it_ thinks that these files moved here.

### DIFF
--- a/plugins/doc_fragments/azure.py
+++ b/plugins/doc_fragments/azure.py
@@ -1,0 +1,113 @@
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2016 Matt Davis, <mdavis@ansible.com>
+# Copyright: (c) 2016 Chris Houseknecht, <house@redhat.com>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+
+class ModuleDocFragment(object):
+
+    # Azure doc fragment
+    DOCUMENTATION = r'''
+
+options:
+    ad_user:
+        description:
+            - Active Directory username. Use when authenticating with an Active Directory user rather than service
+              principal.
+        type: str
+    password:
+        description:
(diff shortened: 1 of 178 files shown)


Fred-sun:
Update document, ready_for_review

PR #124, opened 2020-05-19
Add ephemeral OS disk for azure_rm_virtualmachine

Fred-sun:
##### SUMMARY

Add ephemeral OS disk for azure_rm_virtualmachine.,

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_virtualmachine
##### ADDITIONAL INFORMATION

```paste below

```


plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -843,6 +849,7 @@ def __init__(self):
             vm_identity=dict(type='str', choices=['SystemAssigned']),
             winrm=dict(type='list'),
             boot_diagnostics=dict(type='dict'),
+            ephemeral_os_disk=dict(type='bool', default=False),
```
haiyuazhang: don't set a default value for ephemeral_os_disk.

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1050,6 +1058,13 @@ def exec_module(self, **kwargs):
                 differences = []
                 current_nics = []
                 results = vm_dict
+                current_osdisk = vm_dict['properties']['storageProfile']['osDisk']
+                current_ephemeral = current_osdisk.get('diffDiskSettings',None)
+
+                if self.ephemeral_os_disk and not current_ephemeral:
+                    self.fail('Ephemeral OS disk not updatable: virtual machine ephemeral OS disk is {0}'.format(self.ephemeral_os_disk))
```
haiyuazhang, line 1069: using self.ephemeral_os_disk is not None and self.ephemeral_os_disk != current_ephemeral
as the condition to judge whether self.-ephemeral_os_disk setting is to be updated.

plugins/modules/azure_rm_virtualmachine.py
Fred-sun: Ok! I will remove the default value!

plugins/modules/azure_rm_virtualmachine.py
Fred-sun: Ok! I will updated!

PR #148, opened 2020-06-08
Add ability to remove all Subnet Service Endpoints when supplying an empty list

GerbenWelter:
##### SUMMARY
Previously it was impossible to completely remove Service Endpoints from a subnet because of the way the code was written. Even explicitly supplying an empty list would result in the code not comparing with the current state. This PR adds a minor tweak to fix this scenario and adds a test for it.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
azure_rm_subnet

##### ADDITIONAL INFORMATION
The problem was the way the code evaluated the supplied value for service_endpoints:

```
if self.service_endpoints:
```

When self.service_endpoints is equal to None (default) or '[]', the code will not enter the if-loop. Extending the if-statement as follows will enable to completely remove all Service Endpoints:

```
if self.service_endpoints or self.service_endpoints == []:
```


Fred-sun:
ready_for_review

PR #140, opened 2020-06-01
For Test( 2.9 to 2.10)

Fred-sun:
##### SUMMARY

adding more testing dimention

##### ISSUE TYPE

- Bugfix Pull Request

##### ADDITIONAL INFORMATION

```paste below

```

PR #173, opened 2020-06-24
For Ansible 2.10 compat, change imports to use module_utils already in this collection

AlanCoding:
##### SUMMARY
This collection will not work with Ansible 2.10 because that will no longer host the old `module_utils` specific to content like this.

I verified that imports fail with `ansible==2.10.0a1`.

This solution, using relative imports, will sacrifice compatibility with Ansible 2.8, and I am updating metadata to reflect that. It is still possible to make this work with both 2.8 and 2.10 if you type out the fully qualified collection name for each import, which is like

https://github.com/ansible-collections/azure/blob/8aad2a0c43d16cf0f5619164751e1e7049316f69/plugins/modules/azure_rm_resourcegroup.py#L132

As you can see from this link, some modules already use these fully-qualified style imports. I'm not changing those here.

Why are some modules using fully-qualified imports and others not? I see the one I linked in #45, which states that it is the 4th migration of modules... and I can find #52 migrated a much larger number without using the fully-qualified imports.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils

##### ADDITIONAL INFORMATION
The module utils were already here:

I previously put up #115, and this is very similar, although it's not moving anything, just redirecting stuff.

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -1,0 +1,2 @@
+---
+requires_ansible: '>2.8'
--- a/plugins/modules/azure_rm_lock.py
+++ b/plugins/modules/azure_rm_lock.py
@@ -101,7 +101,7 @@ id:
     sample: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/myResourceGroup/providers/Microsoft.Authorization/locks/keep"
 '''  # NOQA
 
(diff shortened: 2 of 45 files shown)


AlanCoding:
I think @Fred-sun addressed all these in https://github.com/ansible-collections/azure/pull/140

PR #53, opened 2020-03-06
Support azure cli credentials with multiple `subscription_id`s

internetionals:
This is the same change I have a pull request for at the Ansible main repository, but it isn't entirely clear what the correct path for new changes is for the Azure Ansible modules.

See original pull request: ansible/ansible#65331
(and also: Azure/azure_preview_modules#358)

I also couldn't find the doc_fragments part from the original change, so this change doesn't update the documentation, which the original does.

Also relates to these issues/pull requests in the main Ansible repository:

- ansible/ansible#48089
- ansible/ansible#63182

Below is the text from the pull request against the main Ansible repository:

##### SUMMARY
When using credentials obtained from the azure cli only the default subscription can be used, even though the azure cli is authenticated for multiple subscriptions.

This change passes any optionally specified `subscription_id` (similar to how it's done for `auth_source: msi`) along when requesting the azure cli credentials. If none is specified it falls back to the current behaviour of selecting the default subscription.

The only other change that had to be made was in the `auth_source: auto` case where we would always assume that all credential information is passed using arguments when `subscription_id` is set. I changed this to check the fields that actually refer to specific credentials (namely `client_id` (for service principals) and `ad_user` (for user names)). This way we still fall through to the azure cli method, eventhough only `subscription_id` is explicitly set.

Fixes ansible/ansible#63182
Fixes Azure/azure_preview_modules#321

This effectively fixes the same problem as ansible/ansible#48089 is trying to fix, only in a way that is more consistent with how credentials and subscription id's are determined for other `auth_source`s. This was one of the main criticisms that's probably holding that pull request up.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
azure_rm_common

##### ADDITIONAL INFORMATION
We have different Azure subscriptions for our prod and dev environments. By default we only work on the dev environment, but we have a number of playbooks that we want to work with all virtual machines at once.

Demonstration playbook:
```
- name: Get azure vm info from different subscriptions
  hosts: localhost
  tasks:
  - azure_rm_virtualmachine_info:
      name: prod-db
      resource_group: prod
      subscription_id: XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX
      auth_source: cli
  - azure_rm_virtualmachine_info:
      name: dev-db
      resource_group: dev
      subscription_id: YYYYYYYY-YYYY-YYYY-YYYY-YYYYYYYYYYYY
      auth_source: cli
```

Before:
```
[...]

PR #195, opened 2020-07-21
Support azure cli credentials with multiple `subscription_id`s

UnwashedMeme:
##### SUMMARY

This is very similar to #53  but taken a bit further:
* rebased onto `dev` branch 
* Added more documentation
* Tried to cleanup/simply `azure_rm_common` a bit (not very much).
  * move some defaults  out of `__init__` onto class
  * use python kwargs instead fo passing in a `dict(...)`
  * use ansible's `env_fallback` for `auth_source`.

I'm hoping to get a bit further but wanted to put up the WIP and get any feedback.

### DIFF
--- a/plugins/doc_fragments/azure.py
+++ b/plugins/doc_fragments/azure.py
@@ -68,16 +68,19 @@ options:
     auth_source:
         description:
             - Controls the source of the credentials to use for authentication.
-            - If not specified, ANSIBLE_AZURE_AUTH_SOURCE environment variable will be used and default to C(auto) if variable is not defined.
-            - C(auto) will follow the default precedence of module parameters -> environment variables -> default profile in credential file
-              C(~/.azure/credentials).
-            - When set to C(cli), the credentials will be sources from the default Azure CLI profile.
             - Can also be set via the C(ANSIBLE_AZURE_AUTH_SOURCE) environment variable.
+            - When set to C(auto) (the default) the precedence is module parameters -> C(env) -> C(credential_file) -> C(cli).
+            - When set to C(env), the credentials will be read from the environment variables
+            - When set to C(credential_file), it will read the profile from C(~/.azure/credentials).
+            - When set to C(cli), the credentials will be sources from the Azure CLI profile. C(subscription_id) or the environment variable
+              C(AZURE_SUBSCRIPTION_ID) can be used to identify the subscription ID if more than one is present otherwise the default
+              az cli subscription is used.
             - When set to C(msi), the host machine must be an azure resource with an enabled MSI extension. C(subscription_id) or the
               environment variable C(AZURE_SUBSCRIPTION_ID) can be used to identify the subscription ID if the resource is granted
               access to more than one subscription, otherwise the first subscription is chosen.
             - The C(msi) was added in Ansible 2.6.
         type: str
+        default: auto
         choices:
         - auto
         - cli
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1448,7 +1448,7 @@ class AzureRMAuth(object):
             arg_credentials[attribute] = params.get(attribute, None)
 
         if auth_source == 'msi':
-            self.log('Retrieving credenitals from MSI')
+            self.log('Retrieving credentials from MSI')
             return self._get_msi_credentials(subscription_id=params.get('subscription_id'), client_id=params.get('client_id'))
 
         if auth_source == 'cli':
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1275,8 +1275,10 @@ class AzureRMAuth(object):
                           "define a profile in ~/.azure/credentials, or install Azure CLI and log in (`az login`).")
 
         # cert validation mode precedence: module-arg, credential profile, env, "validate"
-        self._cert_validation_mode = cert_validation_mode or self.credentials.get('cert_validation_mode') or \
-            os.environ.get('AZURE_CERT_VALIDATION_MODE') or 'validate'
+        self._cert_validation_mode = cert_validation_mode or \
+            self.credentials.get('cert_validation_mode') or \
+            os.environ.get('AZURE_CERT_VALIDATION_MODE') or \
+            'validate'
 
         if self._cert_validation_mode not in ['validate', 'ignore']:
             self.fail('invalid cert_validation_mode: {0}'.format(self._cert_validation_mode))
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1277,7 +1277,7 @@ class AzureRMAuth(object):
         # cert validation mode precedence: module-arg, credential profile, env, "validate"
         self._cert_validation_mode = cert_validation_mode or \
             self.credentials.get('cert_validation_mode') or \
-            os.environ.get('AZURE_CERT_VALIDATION_MODE') or \
+            self._get_env('cert_validation_mode') or \
             'validate'
 
         if self._cert_validation_mode not in ['validate', 'ignore']:
@@ -1368,6 +1368,10 @@ class AzureRMAuth(object):
     def _default_fail_impl(self, msg, exception=None, **kwargs):
         raise AzureRMAuthException(msg)
 
+    def _get_env(self, module_key, default=None):
+        "Read envvar matching module parameter"
+        return os.environ.get(AZURE_CREDENTIAL_ENV_MAPPING[module_key], default)
+
     def _get_profile(self, profile="default"):
         path = expanduser("~/.azure/credentials")
         try:
@@ -1390,7 +1394,7 @@ class AzureRMAuth(object):
 
     def _get_msi_credentials(self, subscription_id=None, client_id=None, **kwargs):
         credentials = MSIAuthentication(client_id=client_id)
-        subscription_id = subscription_id or os.environ.get(AZURE_CREDENTIAL_ENV_MAPPING['subscription_id'])
+        subscription_id = subscription_id or self._get_env('subscription_id')
         if not subscription_id:
             try:
                 # use the first subscription of the MSI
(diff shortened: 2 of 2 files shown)


Fred-sun:
@UnwashedMeme Recently we have made some related changes to this repo. We need your help to rebase this PR. Once your rebase is completed, I will test and promote the merger of this PR. Thank you!

UnwashedMeme:
@Fred-sun just rebased again.  I'm also happy with Internetionals work in #53. I'm just trying to take it a bit further and do some more cleanups.

I haven't gotten a chance to test all the code paths (msi, profile, cli, sp).  I'm hoping you have some tests in here for that and I've tried to make each change small for easier reviewing.

plugins/doc_fragments/azure.py
```diff
@@ -69,10 +69,12 @@ class ModuleDocFragment(object):
         description:
             - Controls the source of the credentials to use for authentication.
             - If not specified, ANSIBLE_AZURE_AUTH_SOURCE environment variable will be used and default to C(auto) if variable is not defined.
-            - C(auto) will follow the default precedence of module parameters -> environment variables -> default profile in credential file
-              C(~/.azure/credentials).
-            - When set to C(cli), the credentials will be sources from the default Azure CLI profile.
-            - Can also be set via the C(ANSIBLE_AZURE_AUTH_SOURCE) environment variable.
+            - C(auto) will follow the default precedence of: module parameters -> C(env) -> C(credential_file) -> C(cli).
```
Fred-sun: No semicolon in the description statement。 The default value is defined in the module, you need to specify here and define the default value below.

plugins/module_utils/azure_rm_common.py
```diff
@@ -1388,10 +1406,14 @@ def _get_msi_credentials(self, subscription_id_param=None, **kwargs):
             'subscription_id': subscription_id
         }
 
-    def _get_azure_cli_credentials(self, resource=None):
+
```
Fred-sun: There is a line between functions (there are two blank lines)。

plugins/module_utils/azure_rm_common.py
```diff
@@ -1415,30 +1437,26 @@ def _get_env_credentials(self):
 
         return None
 
-    # TODO: use explicit kwargs instead of intermediate dict
-    def _get_credentials(self, params):
+
```
Fred-sun: There is a line between functions (there are two blank lines)。

Fred-sun:
@UnwashedMeme  Could you please help answer the following questions?
```e.g.

1. When subscription_ID is specified in the parameter, this parameter has no effect. (When both sunbscription_ID and auth_source parameters exist, you do not specify a priority in use)

File: plugins/module_utils/azure_rm_common.py (hunk 2 above)
Comment (UnwashedMeme) [reply plugins/module_utils/azure_rm_common.py]: fixed

File: plugins/module_utils/azure_rm_common.py (hunk 3 above)
Comment (UnwashedMeme) [reply plugins/module_utils/azure_rm_common.py]: fixed

File: plugins/doc_fragments/azure.py (hunk 1 above)
Comment (UnwashedMeme) [reply plugins/doc_fragments/azure.py]: Please let me restate your request to ensure I understand it.

* I should change this line to remove everything after the colon so that it reads `C(auto) will follow the module default precedence`
* I should update the `notes` section below to define the default precedence order

Is this correct?

**Comment by UnwashedMeme** [comment]
> @UnwashedMeme Could you please help answer the following questions?
> 
>> 1. When subscription_ID is specified in the parameter, this parameter has no effect. (When both sunbscription_ID and auth_source parameters exist, you do not specify a priority in use)

I'm afraid I don't understand the question. I'm going to restate it a bit differently and try to answer that. if I miss something, will you please clarify?

> When the module parameters include `subscription_id` and `auth_source` what happens?

The `auth_source` previously and still does determine what authentication source will be used. The only time this isn't true is if `auth_source` is `auto` or not specified (default to `auto`). The docs already declare it "controls" which mechanism is used.

This PR reduces the number of cases in which `subscription_id` is unexpectedly ignored.

`subscription_id` is already common parameter for at least 2 authentication sources: service principal and [msi](https://github.com/ansible-collections/azure/blob/0328969766e5fc2dd17acc1f8a160279ce99457c/plugins/module_utils/azure_rm_common.py#L1431).

I don't believe the docs currently say that `subscription_id` implies "only use service principal" in any of [subscription_id](https://github.com/ansible-collections/azure/blob/0328969766e5fc2dd17acc1f8a160279ce99457c/plugins/doc_fragments/azure.py#L31), [auth_source](https://github.com/ansible-collections/azure/blob/0328969766e5fc2dd17acc1f8a160279ce99457c/plugins/doc_fragments/azure.py#L68), or [notes](https://github.com/ansible-collections/azure/blob/0328969766e5fc2dd17acc1f8a160279ce99457c/plugins/doc_fragments/azure.py#L103). The closest they get is "To authenticate via service principal, pass subscription_id, client_id, secret and tenant."  To me that does not mean that `subscrition_id` _only_ belongs to that set, especially when the docs also say, under `msi`, that "`subscription_id` or [envvar] can be used to identify the subscription ... if ...more than one"

That current behavior of tying `subscription_id` to service principal is from [this `if` statement](https://github.com/ansible-collections/azure/blob/0328969766e5fc2dd17acc1f8a160279ce99457c/plugins/module_utils/azure_rm_common.py#L1464). I don't believe this was a good test because it is possible in the Azure api to authenticate with a service principal specifying the the client id, secret, and tenant without `subscription_id`.  This test on L1464 doesn't ensure any of those important params are actually present. 

The change on [L480](https://github.com/ansible-collections/azure/pull/195/files#diff-d7197e04dc3e55d533b0b1e35f2d413cR1480) of this PR improves this situation as it requires either `client_id` or `ad_user` to be present in the module params; both of those are better distinguishers of an authentication mechanism than `subscription_id`.

**Comment by Fred-sun** [comment]
@UnwashedMeme I have been busy with other more urgent things recently, I will watch this later. Thank you!

**Comment by UnwashedMeme** [comment]
rebased; @Fred-sun  have you gotten a chance to look at this yet?

File: plugins/doc_fragments/azure.py (hunk 1 above)
Comment (Fred-sun) [reply plugins/doc_fragments/azure.py]: yes! Thanks for your update!

Comment (Fred-sun) [inline plugins/doc_fragments/azure.py]: @UnwashedMeme Please added the default value for this parameter, such as 
```e.g
default: auto

Comment (Fred-sun) [inline plugins/doc_fragments/azure.py]:
```suggestion
            - C(auto) will follow the default precedence is module parameters -> C(env) -> C(credential_file) -> C(cli).
```

Fred-sun: remove ':'

Fred-sun (commented):
@UnwashedMeme  All task test PASS, but sanity fail, please update as follow comment and rebase again! Thank you very much!

plugins/doc_fragments/azure.py
UnwashedMeme: I tried to follow your suggestions as well as to make the format of this entire section more consistent, "when set to _ ...". Let me know what you think.

UnwashedMeme: done

Fred-sun:
> @UnwashedMeme Could you please help answer the following questions?
> 
> ```
> 
> 1. When subscription_ID is specified in the parameter, this parameter has no effect. (When both sunbscription_ID and auth_source parameters exist, you do not specify a priority in use)
> ```

It's my mistake!  Cancel this comment!

plugins/module_utils/azure_rm_common.py
```diff
@@ -1391,10 +1409,13 @@ def _get_msi_credentials(self, subscription_id_param=None, **kwargs):
             'subscription_id': subscription_id
         }
 
-    def _get_azure_cli_credentials(self, resource=None):
+    def _get_azure_cli_credentials(self, subscription_id=None, resource=None):
         if self.is_ad_resource:
             resource = 'https://graph.windows.net/'
-        credentials, subscription_id = get_azure_cli_credentials(resource)
+        subscription_id = subscription_id or self._get_env('subscription_id')
```
Fred-sun, line 1415: @UnwashedMeme Why do you catch the subscription_id of the environment variable here? auth_source='cli' uses the credentials of 'az loggin', so the environment variable subscription should not be used in this function, please update. Thank you!

UnwashedMeme: > why do you catch the subscription_id of the environment variable here?

Because that is a valid way of providing data to these modules and this PR reduces the number of cases in which a user provided subscription_id is unexpectedly ignored.

More broadly *"The authentication mechanism (service principal, msi, cli) and parameter passing convention (module parameters, environment variables, profile file) should be as loosely coupled as possible."* 

I believe you have been looking at this as "There currently several distinct parameter sets that are entirely separate depending upon which `auth_source` is specified."  I don't believe that is true in existing code except by accident and it certainly isn't stated in the module documentation. 

The only places subscription_id should be required is when we don't have a mechanism for looking it up associated with the authentication.   The [service principal authentication code](https://github.com/ansible-collections/azure/blob/dev/plugins/module_utils/azure_rm_common.py#L1315) does not use the subscription_id. But since we need to have a subscription_id [when creating the mgmt client](https://github.com/ansible-collections/azure/blob/dev/plugins/module_utils/azure_rm_common.py#L851) it is required.  When not set explicitly, some mechanisms like MSI and CLI can provide a default. _If the module user provides a subscription id through any mechanism it should be respected._

Fred-sun: Yes, but if you assign the subscription_id of the environment variable here, then auth_source= cli will not use the credentials of 'az logging', which is contrary to the designed auth_source parameter.

Fred-sun: So I'm think here, if you set ’auth_source= cli‘， then you're going to use the 'az login' credentials, and if you can't catch the credentials or if the credentials are wrong, you're going to throw an exception,' az Login 'credentials have a problem. Thank you very much!

UnwashedMeme: >  if you assign the subscription_id of the environment variable here, then auth_source= cli will not use the credentials of 'az logging'

This is not a problem because `subscription_id` is not part of the credentials from `az login`.  Please recall that the credentials from az login can grant access to many subscriptions. The subscription_id module parameter or environment variable just allows the user to specify which one this module should actually target.

As with _any_ auth_source, if the user specifies a subscription_id that the credentials aren't authorized for it won't work.

Fred-sun: Yes, I'll also update my description, the subscription_id here should get the subscription_id of the playbook parameter, not the one in the environment variable. Thank you very much!

PR #196, opened 2020-07-21
Several changes related to multiple IP configs (ipv4 + ipv6)

aardbol:
##### SUMMARY

**azure_rm_networkinterface.py**
We should make the primary option required because if it's not set, the default value is always "no", while Azure's default value is "yes" when ip configurations total == 1 and "no" when ip configurations total > 1. This causes the state of the task to be "changed" on every run (and make it not seem idempotent). Requiring this option to be set, will fix this issue and remove any confusion.

**azure_rm_virtualnetwork_info**
Return all address prefixes if both an IPv4 and IPv6 prefix have been set

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request

##### COMPONENT NAME

azure_rm_networkinterface.py
azure_rm_virtualnetwork_info

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_networkinterface.py
+++ b/plugins/modules/azure_rm_networkinterface.py
@@ -156,7 +156,7 @@ options:
                 description:
                     - Whether the IP configuration is the primary one in the list.
                 type: bool
-                default: 'no'
+                required: true
             application_security_groups:
                 description:
                     - List of application security groups in which the IP configuration is included.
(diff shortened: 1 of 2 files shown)


haiyuazhang:
@aardbol would you mind updating the test case as well? It's in tests/integration/targets/azure_rm_networkinterface/tasks/main.yml

aardbol:
> @aardbol would you mind updating the test case as well? It's in tests/integration/targets/azure_rm_networkinterface/tasks/main.yml

I'd rather not though. I've already moved on to Terraform because there were many things missing in this module. Feel free to close it or make the last modification yourself.

haiyuazhang:
author has found a alternative solution

PR #197, opened 2020-07-21
Improve OS detection when VM has no osProfile

TiTi:
We have VM without OS profile (checked on https://resources.azure.com/) and they appear as "unknown" os :-/

This appear to be a known issue: https://support.microsoft.com/en-ph/help/4018140/computer-names-of-specialized-virtual-machines-are-missing-or-blank-in

##### SUMMARY
Improve OS detection when VM has no osProfile

We have VM without OS profile (checked on https://resources.azure.com/) and they appear as "unknown" os :-/

This appear to be a known issue: https://support.microsoft.com/en-ph/help/4018140/computer-names-of-specialized-virtual-machines-are-missing-or-blank-in

##### ISSUE TYPE
- Bugfix Pull Request


Fred-sun:
@Thank you for submitting this PR. Can you tell us in detail the purpose of adding these? What if osType is MacOS? Can you use try-catch to include the added code? Thank you!

TiTi:
> @thank you for submitting this PR. Can you tell us in detail the purpose of adding these? What if osType is MacOS? Can you use try-catch to include the added code? Thank you!

Hello,
We have some Azure virtual machines which are detected as having "**unknown**" os, despite being Windows Virtual machines and detected as so in the azure portal.
We troubleshooted this by checking ansible-inventory:
`{os_profile = {'system': 'unknown'}}`
and going to https://resources.azure.com/ to see the description of our VMs and checking against source code of azure_rm.py 
Those that appear as unknown do not have the osProfile:
image
Compared to other/proper vms:
image

We don't really know how they ended up like that but... it's the case.
And we noticed we got the OS type info in the os_disk property:
`|  |  |--{os_disk = {'name': 'VM-NAME-OS', 'operating_system_type': 'windows'}}`

Thus, we made this PR to improve os detection by adding an else case, thus not modifying existing behavior.
There is no MacOS scenario in the if part, and also I don't think it's possible to have MacOS virtual machines on azure, right?

Fred-sun:
> We have VM without OS profile (checked on https://resources.azure.com/) and they appear as "unknown" os :-/
> 
> This appear to be a known issue: https://support.microsoft.com/en-ph/help/4018140/computer-names-of-specialized-virtual-machines-are-missing-or-blank-in
> 
> ##### SUMMARY
> Improve OS detection when VM has no osProfile
> 
> We have VM without OS profile (checked on https://resources.azure.com/) and they appear as "unknown" os :-/
> 
> This appear to be a known issue: https://support.microsoft.com/en-ph/help/4018140/computer-names-of-specialized-virtual-machines-are-missing-or-blank-in
> 
> ##### ISSUE TYPE
> * Bugfix Pull Request

Thanks for your update! I will push for review and merged!

PR #220, opened 2020-08-06
Add execution environment metadata

AlanCoding:
##### SUMMARY
We are building tooling to enable building container images with necessary python & system requirements necessary to use content inside of a collection. The project for this is at:

https://github.com/ansible/ansible-builder

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
meta/

##### ADDITIONAL INFORMATION
It is intended that this doesn't break any `ansible-test` tests.

For the case of this collection, the only thing we need is a pointer to identify where the python requirements are.

The schema is relatively well stabilized, but could always still change depending on feedback.

### DIFF
--- a/meta/execution-environment.yml
+++ b/meta/execution-environment.yml
@@ -1,0 +1,3 @@
+dependencies:
+  python: requirements-azure.txt
+version: 1


howardjones:
Is this information used by anything else? (e.g. awx/tower for pulling in non-ansible dependencies)
I am seeing issues with this collection which seem to be out of date dependencies in awx.

AlanCoding:
These are not used by AWX yet. The current manifest of requirements used by the virtual environment for playbooks in AWX is very old and complicated by things like needing to support python 2. The current requirements are right now only used in experimental builds of the AWX default execution environment, but this will become the way to run jobs in a coming release.

PR #257, opened 2020-09-14
Add support for multi-container apps to azure_rm_webapp

kangasta:
##### SUMMARY

Pass-through container name to applications linux version if it starts with one of the allowed values for multi-container application, `COMPOSE|` or `KUBE|`, in order to allow configuring multi-container applications through existing container settings.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_webapp.py


plugins/modules/azure_rm_webapp.py
```diff
@@ -100,7 +100,7 @@
         suboptions:
             name:
                 description:
-                    - Name of container, for example C(imagename:tag).
+                    - Name of the container, for example C(imagename:tag). To create a multi-container app, the name should be C(COMPOSE|) or C(KUBE|) followed by base64 encoded configuration.
```
Fred-sun: Why use such a naming method? Is this different from normal naming?
```suggestion
                    - Name of the container, for example C(imagename:tag).
                    - To create a multi-container app, the name should be 'COMPOSE|' or 'KUBE|' followed by base64 encoded configuration.
```

kangasta: This is the format that Azure seems to use to differentiate between single and multi-container applications. `az webapp config container show` returns the containers name in this format if webapp is a multi-container application.

Fred-sun: I see! Thank you very much!

plugins/modules/azure_rm_webapp.py
```diff
@@ -234,6 +235,16 @@
...
 
+    - name: Create a multi-container web app
+      azure_rm_webapp:
+        resource_group: myResourceGroup
+        name: myMultiContainerWebapp
+        plan: myAppServicePlan
+        app_settings:
+          testkey: testvalue
+        container_settings:
+          name: "COMPOSE|{{ lookup('file', 'docker-compose.yml') | b64encode }}"
```
Fred-sun, line 246: COMPOSE and DOCKER should be followed by lowercase。
```suggestion
          name: COMPOSE{{ lookup('file', 'docker-compose.yml') | b64decode }}
```

Fred-sun:
kindly ping!

plugins/modules/azure_rm_webapp.py
kangasta: To my understanding, Azure webapp API uses base64 encoded docker-compose file content with `COMPOSE|` prefix as the container name when deploying multi-container application. For example: https://docs.microsoft.com/en-us/azure/app-service/tutorial-multi-container-app#update-app-with-new-configuration

PR #368, opened 2020-12-16
converted computer_name to use dictionary .get method.

### LINKED ISSUE #367: 1.3.0 broke our pipelines (no osProfile)
##### SUMMARY
Hello,
Release 1.3.0 broke our pipeline because the inventory could not be loaded:
`WARNING]: Unable to parse /mnt/data/...our_path.../inventory.azure_rm.yml as an inventory source`

Fortunately I think I found the issue. It's related to #197

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
` azure/plugins/inventory/azure_rm.py`

##### ANSIBLE VERSION

```paste below
2.10.4
```

##### ANALYSIS

There are 2 commits in azure/plugins/inventory/azure_rm.py between 1.2.0 and 1.3.0  
1 is documentation and the other is [1436f7937eeafc2f26743b2fc383b549f8f99d48](https://github.com/ansible-collections/azure/commit/1436f7937eeafc2f26743b2fc383b549f8f99d48)

It seems harmless but then I remember the PR #197 I made a while ago which was merged:  
We have some azure VM without OS profile ! See screenshots in the PR.

Thus, I suspect that `self._vm_model['properties']['osProfile']` is None
Therefore `.get('computerName')` likely cause an exception.

=> that's why inventory is not loaded

Could you fix this please?
Graceful degradation is acceptable (set computerName to empty string / None)

Thanks !

For now we have fixed to 1.2.0 in our code

##### EXPECTED RESULTS
Inventory is loaded

##### ACTUAL RESULTS
Inventory not loaded
Works on 1.2.0

paultaiton:
##### SUMMARY

Fixes #367 

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm inventory
##### ADDITIONAL INFORMATION

See #367 for issue. This PR converts the lookup to using the dictionary .get method so that a missing key will default to None.

```paste below

```

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -517,7 +517,7 @@ class AzureHost(object):
             id=self._vm_model['id'],
             location=self._vm_model['location'],
             name=self._vm_model['name'],
-            computer_name=self._vm_model['properties']['osProfile'].get('computerName'),
+            computer_name=self._vm_model['properties'].get('osProfile', {}).get('computerName'),
             availability_zone=av_zone,
             powerstate=self._powerstate,
             provisioning_state=self._vm_model['properties']['provisioningState'].lower(),


paultaiton:
The code change is trivial, but I have asked the bug reporting user @TiTi  , to test the change locally and see if it corrects the problem in their environment.

PR #616, opened 2021-08-26
Update azure_rm_manageddisk Doc to reflect return value

nbr23:
##### SUMMARY

The azure_rm_manageddisk documentation for the module's return value is incorrect. This fixes the documentation to match reality.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

- azure_rm_manageddisk

##### ADDITIONAL INFORMATION

The azure_rm_manageddisk documentation mentions a return dict of this shape:
```
{
 'id': diskid
 'state': disk_state
 'changed': true
}
```

However, the 'id' field is not present in the return value and must be fetched from the 'state' field. This PR removes 'id' as a field from the description, and adds a list of the fields present in 'state'.

The tests already reflect this, no update needed.

### DIFF
--- a/plugins/modules/azure_rm_manageddisk.py
+++ b/plugins/modules/azure_rm_manageddisk.py
@@ -199,10 +199,8 @@ state:
         os_type:
             description:
                 - Type of Operating System.
-            choices:
-                - linux
-                - windows
             type: str
+            sample: linux
         disk_size_gb:
             description:
                 - Size in GB of the managed disk to be created.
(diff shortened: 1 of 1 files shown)


plugins/modules/azure_rm_manageddisk.py
```diff
@@ -163,16 +163,59 @@
...
+            type: str
+            sample: copy
+        source_uri:
+            description:
+                - URI to a valid VHD file to be used or the resource ID of the managed disk to copy.
+            type: str
+        os_type:
+            description:
+                - Type of Operating System.
+            choices:
```
Fred-sun: @nbr23 The return value should not contain the choices option, it should have the sample option. Thank you 1

nbr23: Thanks, fixed!

xuzhang3:
LGTM

PR #621, opened 2021-08-27
Fix os_type comparison in azure_rm_manageddisk with existing disk

nbr23:
##### SUMMARY

Fix os_type comparison in azure_rm_manageddisk always returning true.

The current comparison was happening between a string (eg `linux`) and a OperatingSystemTypes variable (eg `OperatingSystemTypes.linux`), always returning `True`.
This impacts the value of the `changed` variable, signaling a change when none was done.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- azure_rm_manageddisk

### DIFF
--- a/plugins/modules/azure_rm_manageddisk.py
+++ b/plugins/modules/azure_rm_manageddisk.py
@@ -442,7 +442,7 @@ class AzureRMManagedDisk(AzureRMModuleBase):
             if not found_disk['disk_size_gb'] == new_disk['disk_size_gb']:
                 resp = True
         if new_disk.get('os_type'):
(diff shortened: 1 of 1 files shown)


nbr23:
Bundling https://github.com/ansible-collections/azure/pull/621/commits/d4004edc6cd46950547888aa5aae3110d8838429 in as well, which leverages `self.compute_models.OperatingSystemTypes(STR)` to get the `OperatingSystemTypes` object instead of the dict base lookup.
Not a game changer but it homogenizes things, and future proofs (if another OS ends up being supported...)

Cheers!

xuzhang3:
LGTM

PR #627, opened 2021-09-02
azure_rm_availabilityset:  fix checkmode support

nbr23:
##### SUMMARY

- Fix "changed" status in check_mode when creating / updating resource (changed used to always return "false", now it returns true if resources need to be created / updated).
- Fix check_mode for `state: absent`: currently it does NOT respect check_mode if state is `absent`. As a result, the resource will get deleted even in check_mode.
- Fix `state: absent` always returning `changed: true` even if no resource was deleted.

Several new integration tests have been added to validate these cases.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

- azure_rm_availabilityset


xuzhang3:
LGTM

PR #649, opened 2021-10-12
Fix manageddisk unmount documentation

nbr23:
##### SUMMARY

Fix manageddisk unmount documentation.
To unmount a disk, `managed_by` needs to be set to `''`. Simply not setting `managed_by` does not accomplish the desired effect of unmounting.

##### ISSUE TYPE

- Docs Pull Request

##### COMPONENT NAME

azure_rm_manageddisk


xuzhang3:
LGTM

PR #860, opened 2022-05-24
Add new module for Azure NAT gateway

### LINKED ISSUE #794: Add Support for NAT Gateways
##### SUMMARY
Addition of support to create/maintain/delete NAT Gateways, including ability to associate a subnet to a NAT Gateway.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_natgateway

##### ADDITIONAL INFORMATION
In order to have control over the egress IP from Subnets within Azure it is helpful to use NAT Gateways and associate them to one or more subnets.  This would require a new module to support the creation/maintenance of NAT Gateways as well as extension of the Subnet module to support associations to NAT Gateways.

In regions where Availability zones are supported, support for these would be ideal.

Example: Create a Public IP, NAT Gateway using the public IP as the outbound IP, and associate it to a subnet.

```yaml
- name: Create a public ip address
  azure_rm_publicipaddress:
    resource_group: myResourceGroup
    name: myPublicNatIP
    allocation_method: static
    domain_name: foobar
- name: Create a NAT Gateway
  azure_rm_natgateway:
    resource_group: myResourceGroup
    name: myNatGateway
    outbound_public_ip: myPublicNatIP
    location: eastus
    availability_zone: zone1
    idle_timeout_minutes: 10
- name: Create a subnet
  azure_rm_subnet:
    resource_group: myResourceGroup
    virtual_network_name: myVirtualNetwork
    name: mySubnet
    address_prefix_cidr: "10.1.0.0/24"
```

andreadecorte:
##### SUMMARY
Gateway association with subnets to be handled in the subnet instance.
Add relevant integration tests.
I added a default value for SKU for consistency. REST API doesn't have a default value.
I took inspiration from the existing modules for handling of resources IDs.
Support for Public IP prefixes is not implemented, it can be done in a second time.

Fixes #794 
##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
Azure NAT Gataway

### DIFF
--- a/plugins/modules/azure_rm_natgateway.py
+++ b/plugins/modules/azure_rm_natgateway.py
@@ -1,0 +1,414 @@
+#!/usr/bin/python
+#
+# Copyright (c) 2022 Andrea Decorte, <adecorte@redhat.com>
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = """
+---
+module: azure_rm_natgateway
+short_description: Manage Azure NAT Gateway instance
+description:
+    - Create, update and delete instances of Azure NAT Gateway.
+
+options:
+    resource_group:
+        description:
+            - The name of the resource group.
+        required: True
+        type: str
+    name:
+        description:
+            - The name of the NAT Gateway.
+        required: True
+        type: str
+    location:
+        description:
+            - Resource location. If not set, location from the resource group will be used as default.
+        type: str
+    zones:
+        description:
+            - List of Availability Zones in which this NAT Gateway should be located.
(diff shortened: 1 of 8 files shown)


alfwed:
Many thanks for this PR ! 
It would be great to also be able to specify the Public IP to associate with the gateway. My company is using a naming convention for every resource and although we badly need this module, we wouldn't be able to use it as it is because of the lack of the public IP option. Any chance you could add it too?

andreadecorte:
> 

This is the case, you will be able to specify one or more existing public IP addresses as parameter for the NAT gateway. Unless you meant to use  public IP prefixes, and this is indeed not included, as it required an additional new module, but it can come later

alfwed:
Sorry I wasn't looking at the right module file. It looks perfectly fine to me :+1:

xuzhang3:
@andreadecorte LGTM

PR #865, opened 2022-05-26
Changed aad_profile.admin_group_object_ids to take a list of string

Check-8:
##### SUMMARY
Azure expect a list of strings as a value for admin_group_object_ids, while current implementations sends a string.
This causes the following error: `msrest.exceptions.SerializationError: Refuse str type as a valid iter type.`

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
azure_rm_aks


xuzhang3:
@Check-8 LGTM

PR #909, opened 2022-07-01
Add multi task support to attach manageddisk on VM

Fred-sun:
##### SUMMARY

Add multi task support to attach manageddisk on VM

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME

azure_rm_manageddisk.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_manageddisk.py
+++ b/plugins/modules/azure_rm_manageddisk.py
@@ -460,11 +460,22 @@ class AzureRMManagedDisk(AzureRMModuleBase):
             if not self.check_mode:
                 cpu_count = multiprocessing.cpu_count()
                 executor = ThreadPoolExecutor(max_workers=cpu_count)
+                task_result = []
                 for vm_item in self.managed_by_extended:
                     vm_name_id = self.compute_client.virtual_machines.get(vm_item['resource_group'], vm_item['name'])
                     if result['managed_by_extended'] is None or vm_name_id.id not in result['managed_by_extended']:
                         changed = True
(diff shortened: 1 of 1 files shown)


archoversight:
When does this thread get `join()`'ed back to the main thread so that it is properly waited on before the main thread finishes?

How are errors that happen in the thread bubbled back up to the consumer (for example the attach fails)?

What happens if Ansible exits and the main thread exits, the other threads will get torn down in the middle of doing their task? Adding proper async support would alleviate the above concerns.

plugins/modules/azure_rm_manageddisk.py
```diff
@@ -460,7 +461,7 @@ def exec_module(self, **kwargs):
                     vm_name_id = self.compute_client.virtual_machines.get(vm_item['resource_group'], vm_item['name'])
                     if result['managed_by_extended'] is None or vm_name_id.id not in result['managed_by_extended']:
                         changed = True
-                        self.attach(vm_item['resource_group'], vm_item['name'], result)
+                        threading.Thread(target=self.attach, args=(vm_item['resource_group'], vm_item['name'], result)).start()
```
archoversight: Since this gets sent into a new thread, the call to `self.get_managed_disk()` on the next line may run before the thread finishes attaching the disks and it might fail because the new disk is not yet attached.

xuzhang3: default thread has changed to thread executor

xuzhang3:
LGTM

PR #922, opened 2022-07-19
Add new module `azure_rm_account_info` to get facts for current logged in user

mandar242:
##### SUMMARY

Add new module `azure_rm_account_info`.
Provides output equivalent to `az account show`.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

azure_rm_account_info
##### ADDITIONAL INFORMATION

Makes use of below to create a output containing details similar to and of similar format to `az account show`
- https://docs.microsoft.com/en-us/python/api/azure-mgmt-resource/azure.mgmt.resource.subscriptions.v2019_11_01.models.subscription?view=azure-python
- https://docs.microsoft.com/en-us/python/api/azure-graphrbac/azure.graphrbac.operations.signedinuseroperations?view=azure-python#azure-graphrbac-operations-signedinuseroperations-get

### DIFF
--- a/plugins/modules/azure_rm_account_info.py
+++ b/plugins/modules/azure_rm_account_info.py
@@ -1,0 +1,179 @@
+#!/usr/bin/python
+#
+# Copyright (c) 2020 Paul Aiton, < @paultaiton >
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: azure_rm_account_info
+
+version_added: "1.2.0"
+
+short_description: Get Azure Account facts (output of az account show)
+
+description:
+    - Get facts for a specific account or all accounts.
+
+options:
+    id:
+        description:
+            - Limit results to a specific subscription by id.
+            - Mutually exclusive with I(name).
+        type: str
+    name:
+        description:
+            - Limit results to a specific subscription by name.
+            - Mutually exclusive with I(id).
+        aliases:
+            - subscription_name
+        type: str
+    all:
+        description:
+            - If true, will return all subscriptions.
+            - If false will omit disabled subscriptions (default).
+            - Option has no effect when searching by id or name, and will be silently ignored.
+        type: bool
+    tags:
+        description:
+            - Limit results by providing a list of tags. Format tags as 'key:value'.
+            - Option has no effect when searching by id or name, and will be silently ignored.
+        type: list
+        elements: str
+
+extends_documentation_fragment:
+    - azure.azcollection.azure
+
+author:
+    - Paul Aiton (@paultaiton)
+'''
+
+EXAMPLES = '''
+- name: Get facts for one subscription by id
+  azure_rm_subscription_info:
+    id: 00000000-0000-0000-0000-000000000000
+
+- name: Get facts for one subscription by name
+  azure_rm_subscription_info:
+    name: "my-subscription"
+
+- name: Get facts for all subscriptions, including ones that are disabled.
+  azure_rm_subscription_info:
+    all: True
+
+- name: Get facts for subscriptions containing tags provided.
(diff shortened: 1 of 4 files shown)


plugins/modules/azure_rm_account_info.py
```diff
@@ -0,0 +1,229 @@
...
+
+        # Makes use of azure.mgmt.subsctiption.operations.subscriptionoperations
+        # https://docs.microsoft.com/en-us/python/api/azure-mgmt-subscription/azure.mgmt.subscription.operations.subscriptionsoperations?view=azure-python#methods
+
+        try:
+            subscription_list_response = list(self.subscription_client.subscriptions.list())
+        except CloudError as exc:
+            self.fail("Failed to list all subscriptions - {0}".format(str(exc)))
+
+        results['id'] = subscription_list_response[0].subscription_id
```
abikouo, line 176: is this a choice to return only the first subscription ``[0]`` for the account?

plugins/modules/azure_rm_account_info.py
```diff
@@ -0,0 +1,229 @@
...
+        return results
+
+    def get_managed_by_tenants_list(self, object_list):
+
+        result = []
+
+        for item in object_list:
+            result.append({"tenantId": item.tenant_id})
+
+        return result
```
abikouo:
```suggestion
        return [dict(tenantId=item.tenant_id) for item in object_list]
```

abikouo (changes requested):
@mandar242 Thanks for this pull request.
Could you please add some integration tests for this module?

plugins/modules/azure_rm_account_info.py
mandar242: Looking at the documentation and based on my trial-errors for `az account show` here: https://docs.microsoft.com/en-us/cli/azure/account?view=azure-cli-latest#az-account-show
I felt that it returns a single subscription, so adding `[0]`th subscription.
I could be wrong here though. Thoughts?

tests/integration/targets/azure_rm_account_info/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+
+- name: Get facts for current logged in user
+  azure.azcollection.azure_rm_account_info:
+  register: result
+
+- assert:
+    that:
+      - result is not changed
+      - result is not failed
```
abikouo, line 8: It should be good to test that the result contains ``account_info`` and the all expected key from account_info, ``tenantId``, ``state``, etc

abikouo (changes requested):

tests/integration/targets/azure_rm_account_info/tasks/main.yml
mandar242: Added.

tests/integration/targets/azure_rm_account_info/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+      - "'account_info' in result"
+      - "'environmentName' in result.account_info"
+      - "'homeTenantId' in result.account_info"
+      - "'id' in result.account_info"
+      - "'managedByTenants  ' in result.account_info"
```
Fred-sun:
```suggestion
      - "'managedByTenants' in result.account_info"
```

Fred-sun (commented):
Consider deleting lines in the module that are not useful for modifying comments. Thank you very much!

tests/integration/targets/azure_rm_account_info/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
+---
```
Fred-sun:
```suggestion
```

tests/integration/targets/azure_rm_account_info/tasks/main.yml
```diff
@@ -0,0 +1,19 @@
...
+
```
Fred-sun:
```suggestion
```

xuzhang3:
LGTM

PR #966, opened 2022-08-30
Add ability to upgrade nodepool kubernetes version

dhageman:
##### SUMMARY
The current version of azure_rm_aks.py can only upgrade the control plane on an Azure Kubernetes Service.  This PR adds a parameter under the node pool profile called "orchestrator_version" that can be used to activate a kubernetes upgrade for the node pool.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
azure_rm_aks.py

##### ADDITIONAL INFORMATION

The control plane is upgraded using the kubernetes_version on the module.
The node pool is upgraded using orchestrator_version in the agent_pool_profiles.

### DIFF
--- a/plugins/modules/azure_rm_aks.py
+++ b/plugins/modules/azure_rm_aks.py
@@ -109,6 +109,10 @@ options:
                     - 'System'
                     - 'User'
                 type: str
+            orchestrator_version:
+                description:
+                    - Version of kubernetes running on the node pool.
+                type: str
             node_labels:
                 description:
                     - Agent pool node labels to be persisted across all nodes in agent pool.
@@ -549,6 +553,7 @@ def create_agent_pool_profiles_dict(agentpoolprofiles):
         os_type=profile.os_type,
         type=profile.type,
         mode=profile.mode,
+        orchestrator_version=profile.orchestrator_version,
         enable_auto_scaling=profile.enable_auto_scaling,
         max_count=profile.max_count,
         node_labels=profile.node_labels,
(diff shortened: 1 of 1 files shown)


plugins/modules/azure_rm_aks.py
```diff
@@ -605,8 +610,9 @@ def create_addon_profiles_spec():
     vnet_subnet_id=dict(type='str'),
     availability_zones=dict(type='list', elements='int', choices=[1, 2, 3]),
     os_type=dict(type='str', choices=['Linux', 'Windows']),
+    orchestrator_version=dict(type='str', required=False),
     type=dict(type='str', choice=['VirtualMachineScaleSets', 'AvailabilitySet']),
-    mode=dict(type='str', choice=['System', 'User'], requried=True),
+    mode=dict(type='str', choice=['System', 'User'], required=True),
```
Fred-sun:
```suggestion
    mode=dict(type='str', choice=['System', 'User']),
```

dhageman: Hey @Fred-sun - This was just a typo that was preexisting in the module.  I just corrected the typo.  Is it a required option?  I am okay with your suggestion if you feel it is safe to move forward with.

Fred-sun: @dhageman This parameter is not required! Thank you very much!

Fred-sun:
@dhageman I tested this parameter, and it seems that this "orchestrator_version" is consistent with"kubernetes_version". Would you like to try it? 
Thank you very much!

dhageman:
@Fred-sun - Setting a new value for kubernetes_version will only upgrade the control plane of a cluster.  The node pool supporting the cluster will not be upgraded at all which is why this modification is needed.  I imagine that most people will just set both to be the same version when doing upgrades but the flexibility of doing upgrades individually may be nice for some.

xuzhang3:
@dhageman LGTM 🚢

xuzhang3 (approved):

PR #1130, opened 2023-03-29
add  include_host_filters to support the hosts filter

xu001186:
Add a new option "include_host_filters" to the azure_rm inventory plugin to support positive filtering. The reason to add this feature is that currently the azure_rm inventory plugin only supports negative filtering (exclude_host_filters), making it difficult to include hosts based on some complex conditions.

### DIFF
--- a/plugins/modules/azure_rm_deployment.py
+++ b/plugins/modules/azure_rm_deployment.py
@@ -84,7 +84,7 @@ options:
         description:
             - If I(state=present), template will be created.
             - If I(state=present) and deployment exists, it will be updated.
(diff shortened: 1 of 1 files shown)


Fred-sun:
@xu001186 Thank you very much for your contribution, can you help fix the conflict in PR? We'll move forward with incorporating this PR as soon as possible, and 'include_host_filters' is a great parameter to use.

Fred-sun:
@xu001186 Will you continue modify this PR? If you continue, you are very welcome to submit a new commit, thank you!！

PR #1214, opened 2023-07-14
Change the dynamically obtained public IP address to a list

Fred-sun:
##### SUMMARY

Change the dynamically obtained public IP address to a list

##### ISSUE TYPE

- Bugfix Pull Request
- Docs Pull Request
- Feature Pull Request
- New Module Pull Request

##### COMPONENT NAME

azure_rm.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -535,6 +535,8 @@ class AzureHost(object):
             security_group_id=[],
             security_group=[],
             public_ipv4_addresses=[],
+            public_ip_id=[],
+            public_ip_name=[],
             public_dns_hostnames=[],
             private_ipv4_addresses=[],
             id=self._vm_model['id'],
@@ -569,10 +571,10 @@ class AzureHost(object):
                     new_hostvars['private_ipv4_addresses'].append(private_ip)
                 pip_id = ipc['properties'].get('publicIPAddress', {}).get('id')
                 if pip_id:
-                    new_hostvars['public_ip_id'] = pip_id
+                    new_hostvars['public_ip_id'].append(pip_id)
 
                     pip = nic.public_ips[pip_id]
-                    new_hostvars['public_ip_name'] = pip._pip_model['name']
+                    new_hostvars['public_ip_name'].append(pip._pip_model['name'])
                     new_hostvars['public_ipv4_addresses'].append(pip._pip_model['properties'].get('ipAddress', None))
                     pip_fqdn = pip._pip_model['properties'].get('dnsSettings', {}).get('fqdn')
                     if pip_fqdn:
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -566,19 +566,22 @@ class AzureHost(object):
         for nic in sorted(self.nics, key=lambda n: n.is_primary, reverse=True):
             # and from the primary IP config per NIC first
             for ipc in sorted(nic._nic_model['properties']['ipConfigurations'], key=lambda i: i['properties'].get('primary', False), reverse=True):
-                private_ip = ipc['properties'].get('privateIPAddress')
-                if private_ip:
-                    new_hostvars['private_ipv4_addresses'].append(private_ip)
-                pip_id = ipc['properties'].get('publicIPAddress', {}).get('id')
-                if pip_id:
-                    new_hostvars['public_ip_id'].append(pip_id)
+                try:
+                    private_ip = ipc['properties'].get('privateIPAddress')
+                    if private_ip:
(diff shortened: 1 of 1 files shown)


xuzhang3 (approved):

epopisces:
Looks like there may be an bug in the chain() function, when there is no public IP address instead of simply returning the private IP address an index out of range error causes the entire execution to fail.  Unfortunately this change breaks our dynamic inventories, because most of our hosts do not have a public IP address assigned.

The error we now get (with emphasis in bold):

```python
ansible-inventory [core 2.15.8]
  config file = /runner/project/ansible.cfg
  configured module search path = ['/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/lib/python3.9/site-packages/ansible
  ansible collection location = /runner/requirements_collections:/runner/.ansible/collections:/usr/share/ansible/collections:/usr/share/automation-controller/collections
  executable location = /usr/local/bin/ansible-inventory
  python version = 3.9.18 (main, Sep  7 2023, 00:00:00) [GCC 11.4.1 20230605 (Red Hat 11.4.1-2)] (/usr/bin/python3)
  jinja version = 3.1.2
  libyaml = True
Using /runner/project/ansible.cfg as config file
redirecting (type: inventory) ansible.builtin.azure_rm to azure.azcollection.azure_rm
[WARNING]:  * Failed to parse
/runner/project/inventories/azure/all_running_hosts.azure_rm.yml with
ansible_collections.azure.azcollection.plugins.inventory.azure_rm plugin: list
index out of range
  File "/usr/local/lib/python3.9/site-packages/ansible/inventory/manager.py", line 293, in parse_source
    plugin.parse(self._inventory, self._loader, source, cache=cache)
  File "/runner/requirements_collections/ansible_collections/azure/azcollection/plugins/inventory/azure_rm.py", line 232, in parse
    self._get_hosts()
  File "/runner/requirements_collections/ansible_collections/azure/azcollection/plugins/inventory/azure_rm.py", line 319, in _get_hosts
    next(chain([h.hostvars['public_ip_address'][0]['ipv4_address']], h.hostvars['private_ipv4_addresses']), None))
[WARNING]: Unable to parse
/runner/project/inventories/azure/all_running_hosts.azure_rm.yml as an
inventory source
ERROR! No inventory was parsed, please check your configuration and options.
```
Will open an Issue.

[Issue 1378](https://github.com/ansible-collections/azure/issues/1378) opened to track the troubleshooting, additional details in that Issue.

Fred-sun:
@epopisces Yes, what was previously returned was a list with a string element, and now it is returned as a list with a dictionary element. Is there really a problem with linking the way before? I will consider how to solve this problem, thank you!

epopisces:
@Fred-sun that is what it appears to be--our dynamic inventories were working previously, but now fail with the above error.  

This line: 
/runner/project/inventories/azure/all_running_hosts.azure_rm.yml with ansible_collections.azure.azcollection.plugins.inventory.azure_rm plugin: list index out of range

Seems to suggest it is this change:
```python
# prior code
next(chain(h.hostvars['public_ipv4_addresses'], h.hostvars['private_ipv4_addresses']), None))

# after the merge
next(chain([h.hostvars['public_ip_address'][0]['ipv4_address']], h.hostvars['private_ipv4_addresses']), None))
```
Trying to use debug and other mechanisms to determine what the value of h.hostvars['public_ip_address'] is during an Inventories > Sources > sync operation, but it is tricky working in AWX vs on a local development workstation (I don't have the same perms that our AWX runners have).  If the value of h.hostvars['public_ip_address'] is equal to "" (empty), then attempting to use h.hostvars['public_ip_address'][0][...] would result in a list index out of range error.  If the value was False, None, or a string then the result would have been either a different error or a single char, for instance.

We have one inventory source (a small Azure subscription) that has a single host with a public IP address--that is the only one for which the dynamic inventory sync process doesn't fail: all the others fail with the above error, but all of those have VMs that do not have public IPs.

Quite possibly this may be related to us using an older version of AWX while also using the latest azure_rm collection, so I would be curious to see if the results were the same on a newer version of AWX.  I went through the release notes since our version up until current and didn't see anything that appeared to be impacting. . .

We are on:
AWX Tower version: 21.4.0

epopisces:
Upgrading to AWX Tower version 23.5.1 (using latest operator) did not resolve, same error occurs (so doesn't appear to be related to the old AWX version specifically).  Which makes sense, the execution environment is separate from the AWX control plane.

Fred-sun:
@epopisces I am very sorry for the trouble caused to you. I have fixed this problem in #1379. Thank you!

epopisces:
Appreciate the fast responses!  I'll watch the merge, and will test and report back once the new version is available.

Issue #1223, opened 2023-07-27
The `subscription_id` Module Parameter Is Used Contradictively


timway:
##### SUMMARY
Related to #1218 but broader. Most if not all `azure.azcollection` modules leverage the `subscription_id` module argument provided by `AzureRMModuleBase` class. It has a fallback to the environment variable `AZURE_SUBSCRIPTION_ID` that is commonly used to inject credentials into playbooks either via `ansible-navigator`, AWX, or Automation Controller.

> https://github.com/ansible-collections/azure/blob/v1.16.0/plugins/module_utils/azure_rm_common.py#L602-L611

In the `parse_resource_to_dict` method provided by `AzureRMModuleBase` it looks only to `subscription_id` provided by `AzureRMModuleBase` which maps back to `self.azure_auth.subscription_id` with `azure_auth` being an instantiation of `AzureRMAuth`. This creates a binding between any module that has code that interacts with `parse_resource_to_dict` have the subscription ID used for authentication added to the resource ID it generates (assumes).

> https://github.com/ansible-collections/azure/blob/v1.16.0/plugins/module_utils/azure_rm_common.py#L913-L915

On the other hand, modules that use `get_mgmt_svc_client` method provided by `AzureRMModuleBase` to get a follow-on management client instantiation. actually look at the module parameter for `subscription_id`. This creates some modules that are able to use `subscription_id` to target a module to a subscription not used for authentication and actually get resources created in that target subscription but follow-on plays that try to find that resource by name fail.

As stated below, my expectation coming to the collection and viewing how `ansible` as a whole functions is that any module parameter provided overrides its fallback method. This is not the case right now and I would disagree with @xuzhang3 opinion in #1218 that separate parameters should be leveraged for this use case. I don't view this bug as an enhancement request but it is possible that fixing it like I propose would have playbook breaking changes so it would warrant a v2.0.0 release in my opinion.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
azure.plugins.module_utils.azure_rm_common.AzureRMModuleBase

##### ANSIBLE VERSION
```paste below
ansible [core 2.15.2]
  config file = None
  configured module search path = ['/Users/abcd/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /private/var/folders/1m/3w2m5k795nq38zc8mfzz6p600000gn/T/tmp.KodK6p6K/ansible-navigator/lib/python3.9/site-packages/ansible
[...]

PR #1226, opened 2023-07-31
Added azure_rm_backupazurevm support for different subscription_id

Fred-sun:
##### SUMMARY

Added azure_rm_backupazurevm support for different subscription_id

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm_common.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1024,7 +1024,10 @@ class AzureRMModuleBase(object):
 
     @property
     def subscription_id(self):
-        return self.azure_auth.subscription_id
+        if self.module.params.get('subscription_id') is not None:
+            return self.module.params.get('subscription_id')
+        else:
+            return self.azure_auth.subscription_id
 
     @property
     def storage_client(self):


timway:
@Fred-sun this addresses #1223. I wonder if we also add this as a fallback method to the module param vs checking for it alone so that the module param can be used reliably by plugin developers? I'll work up some patches in that direction as well.

I feel it's best to have the plug-in developer checking the most reliable place, the module param, and it being as accurate as possible.

Definitely a step forward but not all the way there yet.

timway:
Also the inventory plug-in will need to be checked. It appears to have a reference to self.azure_auth.subscription_id

Fred-sun:
@timway The self.azure_auth.subscription_id used in the plug-in makes sense because this is where we obtain the credentials and then initialize the resource management client with the credentials and their corresponding subscription_id. Thank you!

timway:
> @timway The self.azure_auth.subscription_id used in the plug-in makes sense because this is where we obtain the credentials and then initialize the resource management client with the credentials and their corresponding subscription_id. Thank you!

It's not used in auth. It again should respect the parameter when applicable. Failing to do so requires a credential object or ENV vars for each subscription. This is visible in the Azure docs around authentication. The AzureRMAuth class does discover the first subscription the credential has access to as a convenience if not provided.

PR #1306, opened 2023-10-30
Add os_disk_encryption_set to azure_rm_virtualmachine

ephracis:
##### SUMMARY
Add the parameter `os_disk_encryption_set` to the
`azure_rm_virtualmachine` module, making it possible to specify which DES to use when encrypting the OS disk for a newly created VM.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
azure_rm_virtualmachine

##### ADDITIONAL INFORMATION

This is required when deploying a VM into one landing zone, using an encrypted Compute Gallery Image from another landing zone. If this is not set it will try to use the DES of the image which will fail since it belongs to another LZ.

### DIFF
--- a/plugins/modules/azure_rm_virtualmachine.py
+++ b/plugins/modules/azure_rm_virtualmachine.py
@@ -229,6 +229,10 @@ options:
         description:
             - Size of OS disk in GB.
         type: int
+    os_disk_encryption_set:
+        description:
+            - ID of disk encryption set for OS disk.
+        type: str
     os_type:
         description:
             - Base type of operating system.
@@ -1088,6 +1092,7 @@ class AzureRMVirtualMachine(AzureRMModuleBase):
             storage_blob_name=dict(type='str', aliases=['storage_blob']),
             os_disk_caching=dict(type='str', aliases=['disk_caching'], choices=['ReadOnly', 'ReadWrite']),
             os_disk_size_gb=dict(type='int'),
+            os_disk_encryption_set=dict(type='str'),
             managed_disk_type=dict(type='str', choices=['Standard_LRS', 'StandardSSD_LRS', 'StandardSSD_ZRS', 'Premium_LRS', 'Premium_ZRS', 'UltraSSD_LRS']),
             os_disk_name=dict(type='str'),
             proximity_placement_group=dict(type='dict', options=proximity_placement_group_spec),
@@ -1172,6 +1177,7 @@ class AzureRMVirtualMachine(AzureRMModuleBase):
         self.os_type = None
         self.os_disk_caching = None
         self.os_disk_size_gb = None
+        self.os_disk_encryption_set = None
         self.managed_disk_type = None
         self.os_disk_name = None
         self.proximity_placement_group = None
@@ -1559,6 +1565,14 @@ class AzureRMVirtualMachine(AzureRMModuleBase):
                             vm_dict['os_profile']['linux_configuration']['disable_password_authentication']:
                         self.fail("(PropertyChangeNotAllowed) Changing property 'linuxConfiguration.disablePasswordAuthentication' is not allowed.")
 
+                if self.os_disk_encryption_set is not None and \
+                    vm_dict['storage_profile'].get('os_disk') is not None and \
+                    vm_dict['storage_profile']['os_disk'].get('managed_disk') is not None and \
+                    vm_dict['storage_profile']['os_disk']['managed_disk'].get('disk_encryption_set') is not None:
+                    if self.os_disk_encryption_set != \
+                            vm_dict['storage_profile']['os_disk']['managed_disk']['disk_encryption_set']:
+                        self.fail("(PropertyChangeNotAllowed) Changing property 'storage_profile.os_disk.managed_disk.disk_encryption_set' is not allowed.")
+
                 # Defaults for boot diagnostics
                 if 'diagnostics_profile' not in vm_dict:
                     vm_dict['diagnostics_profile'] = {}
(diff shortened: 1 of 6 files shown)


plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -206,6 +206,9 @@
     os_disk_size_gb:
         description:
             - Size of OS disk in GB.
+    os_disk_encryption_set:
+        description:
+            - ID of disk encryption set for OS disk.
```
Fred-sun, line 234: Missing type
```suggestion
            - ID of disk encryption set for OS disk.
        type: str
```

ephracis:
I added some integration tests to use the new disk encryption which works fine.

tests/integration/targets/azure_rm_virtualmachine/tasks/setup_des.yml
```diff
@@ -0,0 +1,58 @@
+
```
Fred-sun, line 4: delete blank lines
```suggestion
```

tests/integration/targets/azure_rm_virtualmachine/tasks/setup_des.yml
```diff
@@ -0,0 +1,58 @@
...
+    key_url: "{{ results['keys'][0]['kid'] }}"
+
+- name: Create disk encryption set
+  azure_rm_diskencryptionset:
+    resource_group: "{{ resource_group }}"
+    name: "{{ des_name }}"
+    source_vault: "myvault{{ des_name }}"
+    key_url: "{{ key_url }}"
+    state: present
+  register: des_results
```
Fred-sun: No new line character at the end of file
```suggestion
  register: des_results
```

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1559,6 +1565,11 @@ def exec_module(self, **kwargs):
                             vm_dict['os_profile']['linux_configuration']['disable_password_authentication']:
                         self.fail("(PropertyChangeNotAllowed) Changing property 'linuxConfiguration.disablePasswordAuthentication' is not allowed.")
 
+                current_os_des_id = vm_dict['storage_profile'].get('os_disk', {}).get('managed_disk', {}).get('disk_encryption_set', {}).get('id', {})
```
Fred-sun:
```suggestion
                current_os_des_id = vm_dict['storage_profile'].get('os_disk', {}).get('managed_disk', {}).get('disk_encryption_set', {}).get('id', None)
```

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1559,6 +1565,11 @@ def exec_module(self, **kwargs):
...
+                if self.os_disk_encryption_set is not None and current_os_des_id is not None:
```
Fred-sun, line 1571:
```suggestion
                if self.os_disk_encryption_set is not None and current_os_des_id is not None and self.os_disk_encryption_set != current_os_des_id:
```

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1172,6 +1177,7 @@ def __init__(self):
         self.os_type = None
         self.os_disk_caching = None
         self.os_disk_size_gb = None
+        self.os_disk_encryption_set = None
```
Fred-sun, line 1180: Please add required_if = [('os_disk_encryption_set', '*', ['managed_disk_type'])] and add "required_if=required_if" to line 1219

Fred-sun:
> I added some integration tests to use the new disk encryption which works fine.

@ephracis  Yes, I see. But os_disk only works if managed_disk_type is not None（Please check lines 1704 through 1711）, and managed_disk_type is not specified in your use case, so it works fine. Thank you!

plugins/modules/azure_rm_virtualmachine.py
ephracis: Done

tests/integration/targets/azure_rm_virtualmachine/tasks/setup_des.yml
ephracis: Done

tests/integration/targets/azure_rm_virtualmachine/tasks/setup_des.yml
ephracis: Done

plugins/modules/azure_rm_virtualmachine.py
ephracis: Done

plugins/modules/azure_rm_virtualmachine.py
ephracis: Done

ephracis:
> > I added some integration tests to use the new disk encryption which works fine.
> 
> @ephracis Yes, I see. But os_disk only works if managed_disk_type is not None（Please check lines 1704 through 1711）, and managed_disk_type is not specified in your use case, so it works fine. Thank you!

Good catch! I've added `managed_disk_type: Standard_LRS` to that specific test case but then I get this error:

> KeyVaultAccessForbidden. Please grant get, wrap and unwrap key permissions to disk encryption set.

Unfortunately I don't see any Ansible module for doing this. Any suggestions?

ephracis:
Ok, I think I've managed to get the integration tests running. The problem was that I had to add the DES to the access list.

Then I ran into a second issue where I had to enable purge protection in order to use the KV for disk encryption. This means that all KVs will reside deleted but unpurged for 90 days. So in order to not get naming conflicts with these keys, they are given a date-based suffix.

Let me know if there's anything else that needs fixing. 😅

xuzhang3 (approved):

xuzhang3:
@ephracis  LGTM

PR #1309, opened 2023-10-31
Add disk_encryption_set for data disks in azure_rm_virtualmachine

ephracis:
##### SUMMARY
Add the parameter `data_disks.disk_encryption_set` to the `azure_rm_virtualmachine` module, making it possible to specify which DES to use when encrypting the data disk.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
azure_rm_virtualmachine

##### ADDITIONAL INFORMATION
This is required when creating a VM (with data disks) from an encrypted image which resides in another landing zone.

### DIFF
--- a/plugins/modules/azure_rm_virtualmachine.py
+++ b/plugins/modules/azure_rm_virtualmachine.py
@@ -1116,6 +1116,7 @@ class AzureRMVirtualMachine(AzureRMModuleBase):
                 options=dict(
                     lun=dict(type='int', required=True),
                     disk_size_gb=dict(type='int'),
+                    disk_encryption_set=dict(type='str'),
                     managed_disk_type=dict(type='str', choices=['Standard_LRS', 'StandardSSD_LRS',
                                            'StandardSSD_ZRS', 'Premium_LRS', 'Premium_ZRS', 'UltraSSD_LRS']),
                     storage_account_name=dict(type='str'),
--- a/plugins/modules/azure_rm_virtualmachine.py
+++ b/plugins/modules/azure_rm_virtualmachine.py
@@ -262,6 +262,10 @@ options:
                     - Size can be changed only when the virtual machine is deallocated.
                     - Not sure when I(managed_disk_id) defined.
                 type: int
+            disk_encryption_set:
+                description:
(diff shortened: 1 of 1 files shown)


Fred-sun:
@ephracis Please fix the conflict in the document! Thank you very much!

ephracis:
> @ephracis Please fix the conflict in the document! Thank you very much!

Done!

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1865,6 +1869,10 @@ def exec_module(self, **kwargs):
                             else:
                                 data_disk_vhd = None
                                 data_disk_managed_disk = self.compute_models.ManagedDiskParameters(storage_account_type=data_disk['managed_disk_type'])
+                                if data_disk.get('disk_encryption_set'):
```
Fred-sun, line 1873: Does it mean that 'disk_encryption_set' can only be specified if 'managed_disk_type' is not configured for data_disks(self.data_disks.get('managed_disk_type')=None)?

Fred-sun:
@ephracis Please add 'disk_encryption_set' to argument. Thank you very much!
```e.g.
nonexistent-parameter-documented: Argument 'disk_encryption_set' found in data_disks is listed in DOCUMENTATION.options, but not accepted by the module argument_spec

plugins/modules/azure_rm_virtualmachine.py
ephracis: I think so. I'm not an expert at Azure but from what I can see, it is only the class `ManagedDiskParameters` which has the parameter `disk_encryption_set` as per the Azure Python SDK: https://learn.microsoft.com/en-us/python/api/azure-mgmt-compute/azure.mgmt.compute.v2023_07_01.models.manageddiskparameters?view=azure-python

I'm not sure if other types of disk can be encrypted using a DES. That might be the case. But for example my current customer has a use case where they use managed disk and need encryption with CMK. This would solve that use case at least.

xuzhang3 (approved):

xuzhang3:
@ephracis  LGTM  🚢

PR #1310, opened 2023-10-31
Add features to azure_rm_galleryimage

### LINKED ISSUE #1282: Ability to set feature data when creating image definition
##### SUMMARY

Add the `features` field to the `azure_rm_galleryimage` module.

##### ISSUE TYPE

This would allow the user to create an image supporting Trusted Launch, which is done using the CLI like so:

```
az sig image-definition create --features SecurityType=TrustedLaunchSupported ...
```

##### COMPONENT NAME

azure_rm_galleryimage

##### ADDITIONAL INFORMATION

This request comes from a Red Hat customer who requires that all VMs use Trusted Launch and wants to use Ansible to manage their Azure infrastructure. Unfortunately they currently have to use `command` module and the `az` tool, because of this missing feature.

If it is not too much complexity and work, I can create a pull request with this feature. With some guidance.

ephracis:
##### SUMMARY
Add the parameter `features` to the `azure_rm_galleryimage` module

Fixes #1282 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
azure_rm_galleryimage

##### ADDITIONAL INFORMATION
This is required to create VMs with Trusted Launch from these images.

The API version had to be bumped in order to support the properties.features setting.

### DIFF
--- a/plugins/modules/azure_rm_galleryimage.py
+++ b/plugins/modules/azure_rm_galleryimage.py
@@ -162,6 +162,22 @@ options:
                 description:
                     - The product ID.
                 type: str
+    features:
+        description:
+            - A list of gallery image features.
+        type: list
+        elements: dict
+        suboptions:
+            name:
+                description:
+                    - The name of the gallery image feature..
+                type: str
+                required: True
+            value:
+                description:
+                    - The value of the gallery image feature.
+                type: str
+                required: True
     state:
         description:
             - Assert the state of the GalleryImage.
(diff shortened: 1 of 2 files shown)


ephracis:
I've added some integration tests but that suite isn't working because it doesn't use `storage_profile`. This is however fixed in #1311 so if that one is merged, I could rebase this one and it should be green.

xuzhang3 (approved):

xuzhang3:
@ephracis LGTM

PR #1376, opened 2023-12-13
`azure_rm_aduser` - Expose more `password_profile` options: `password_force_change` and `password_force_change_mfa`

pluto00987:
* Normalize 'password' vs 'password_profile' variable and option names
* Add options for 'force password change on next logon'

##### SUMMARY

This adds two additional ansible options mapping to existing options within a user's passwordProfile attribute. Specifically for my use case, I needed to be able to set `forceChangePasswordNextSignIn=False`.

[passwordProfile resource type reference](https://learn.microsoft.com/en-us/graph/api/resources/passwordprofile?view=graph-rest-1.0)

Additionally, it normalizes the option and variable names so that 'password' is a password and 'password_profile' is a PasswordProfile object.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_aduser

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_aduser.py
+++ b/plugins/modules/azure_rm_aduser.py
@@ -75,6 +75,17 @@ options:
             - The password for the user.
             - Used when either creating or updating a user account.
         type: str
+    password_force_change:
+        description:
+            - Whether or not the user will be forced to change their password at next logon.
+            - If unspecified, Azure defaults this to true for new users.
+            - Used when either creating or updating a user account.
+        type: bool
+    password_force_change_mfa:
+        description:
+            - Identical behavior to password_force_change except multi-factor authentication (MFA) must be performed prior to changing the password.
+            - Used when either creating or updating a user account.
+        type: bool
     usage_location:
         description:
             - A two letter country code, ISO standard 3166.
@@ -260,6 +271,8 @@ class AzureRMADUser(AzureRMModuleBase):
             account_enabled=dict(type='bool'),
             display_name=dict(type='str'),
             password_profile=dict(type='str', no_log=True),
+            password_force_change=dict(type='bool', no_log=False),
+            password_force_change_mfa=dict(type='bool', no_log=False),
             mail_nickname=dict(type='str'),
             on_premises_immutable_id=dict(type='str', aliases=['immutable_id']),
             usage_location=dict(type='str'),
@@ -280,6 +293,8 @@ class AzureRMADUser(AzureRMModuleBase):
         self.account_enabled = None
         self.display_name = None
         self.password_profile = None
+        self.password_force_change = None
+        self.password_force_change_mfa = None
         self.mail_nickname = None
         self.on_premises_immutable_id = None
         self.usage_location = None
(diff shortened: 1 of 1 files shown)


Fred-sun:
kindly ping!

Fred-sun:
@pluto00987 Why do you have to redefine parameters that already exist? This has a large impact on the current user -- ‘password_profile'. You are advised not to change the previous parameter names. There is also a slight problem with the documentation. Thank you!

pluto00987:
@Fred-sun 
I opted to change the parameter name because 'password_profile' is a misnomer IMO. I did however leave it as an alias for 'password' which should prevent breaking any existing ansible jobs.
This was my same thought with the code. 'Password' is a thing and 'password profile' is also a thing but the existing variable names were mismatched with what they actually contained.
What issue with the documentation?

Fred-sun:
@pluto00987 This will affect the current user. In this case, please fix the problem described in the above document. As for changing the parameter name, we will discuss it internally. Thank you!

Fred-sun:
@pluto00987 There are no other questions, please make changes to the above part of the document. Thank you!

plugins/modules/azure_rm_aduser.py
```diff
@@ -70,11 +70,24 @@
...
         description:
             - The password for the user.
             - Used when either creating or updating a user account.
         type: str
+        aliases:
+            - password_profile
+    password_force_change:
+        description:
+            - Whether or not the user will be forced to change their password at next logon.
+            - Note: If unspecified, Azure defaults this to true for new users.
```
Fred-sun: DOCUMENTATION.options.password_force_change.description.  Must be a string 
```suggestion
            - If unspecified, Azure defaults this to true for new users.
```

Fred-sun (commented):
Small change requests!

Fred-sun:
kindly ping!

Fred-sun:
Kindly ping!

pluto00987:
Hello Fred-sun.
I applied your change and rebased to current upstream. Please let me know if anything further is needed.
Thank you!

pluto00987:
Fred-sun, I made an error with the first rebase but have corrected it now.

Fred-sun:
@xuzhang3 ready_for_review!

pluto00987:
@Fred-sun I see a problem. The `password` option _is_ in fact already used elsewhere, in azure_rm_common.py. I didn't realize this before but happened to be looking at the module documentation page today, where it is listed. So I can't rename `password_profile` to `password` after all.

So I will need to refactor this patch. I still don't like the name `password_profile` but I suppose leaving the original name alone is easiest. I'm open to other suggestions though. It's unfortunate that whomever implemented the global `ad_user`/`password` options didn't use `ad_password` instead for consistency, but such is life.

Fred-sun:
@pluto00987 Yes, you are right that it should not duplicate the integrated parameter name. After you resubmit, I will review and promote the merger as soon as possible! Thank you!

pluto00987:
Hi @Fred-sun. I've updated the change. Please take a look now. Thank you!

xuzhang3 (approved):

PR #1380, opened 2023-12-15
Allow credential file to not set subscription_id and use the param sp…

is-consulting:
##### SUMMARY
This Pull request allows a user to not set the subscription_id value when using a credential file. This enables use of the "subscription_id" param specified in the dynamic inventory if unset.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
Authentication / Credentials

##### ADDITIONAL INFORMATION

```
# When using a credential File, this parameter was ignored and is now used:
subscription_id: 00000000-0000-0000-0000-0000  
auth_source: auto
include_vm_resource_groups: ['*']
include_vmss_resource_groups: ['*']
default_host_filters: []
conditional_groups:
  linux: "'linux' in os_profile.system"
  windows: "'windows' in os_profile.system"
```

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1514,7 +1514,7 @@ class AzureRMAuth(object):
         "Read envvar matching module parameter"
         return os.environ.get(AZURE_CREDENTIAL_ENV_MAPPING[module_key], default)
 
-    def _get_profile(self, profile="default"):
+    def _get_profile(self, profile="default", subscription_id=None):
         path = expanduser("~/.azure/credentials")
         try:
             config = configparser.ConfigParser()
@@ -1529,8 +1529,10 @@ class AzureRMAuth(object):
             except Exception:
                 pass
 
-        if credentials.get('subscription_id'):
-            return credentials
+        if credentials.get('subscription_id') is None:
+            if subscription_id is not None:
+                credentials['subscription_id'] = subscription_id
+                return credentials
 
         return None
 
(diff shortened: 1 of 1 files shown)


plugins/module_utils/azure_rm_common.py
```diff
@@ -1529,8 +1529,10 @@ def _get_profile(self, profile="default"):
             except Exception:
                 pass
 
-        if credentials.get('subscription_id'):
-            return credentials
+        if credentials.get('subscription_id') is None:
```
xuzhang3, line 1532: The profile will be `None` if `credentials.get('subscription_id')` not `None`.   `credentials.get('subscription_id')` and `subscription_id` which one is the first priority?

is-consulting: In order to not change the behavior I give `credentials.get('subscription_id')` the first priority

xuzhang3: `credentials.get('subscription_id') not None` not handled,  profile will returns as `None`

xuzhang3: @is-consulting  ping  `credentials.get('subscription_id') not None`   not handled. 
```suggestion
        if credentials.get('subscription_id') is None:
            ...
        else:
            return credentials
```

Fred-sun:
@is-consulting kindly ping!

Fred-sun:
This PR is not good because it ignores the issue that the parameter contains other subscritption_id, which is inconsistent with the authenticated subscription_id. When we invoke python-sdk, we need to pass the authenticated subscription_id. Thank you!

Fred-sun:
This issue was modified in #2034. When the subscription_id in the environment variable or credentials is inconsistent with that in playbook, the parameter configuration will be used for the resource ID, and the one in credentials will be used for authentication.

Issue #1431, opened 2024-02-05
Inventory failing when flexible and uniform scalesets in the same RG


Justwmz:
##### SUMMARY
Inventory fails when flexible and uniform scale sets in the same RG.
I can see instances of the flexible scale set if the params "include_vmss_resource_groups" are not assigned.
I think it is because instances in the flexible scale set are created as standalone VMs, so Ansible is able to represent them in the inventory output.
But when I add back include_vmss_resource_groups, inventory is falling.
It seems to me like ansible is not able to parse a list of the instances from the flexible scale set since they have a different machine type.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
azure.azcollection

##### ANSIBLE VERSION
2.16.0

##### COLLECTION VERSION
2.1.1

##### STEPS TO REPRODUCE
Create two types of scale sets in the same RG.
In the inventory assign the params below:
include_vm_resource_groups:
 `- name of the RG`
include_vmss_resource_groups:
 `- name of the RG`

##### EXPECTED RESULTS
Inventory output should present instances from both types of scale sets.

Fred-sun:
@Justwmz Obtaining the unifrom VMSS instance fails because the attribute ids of flexible VMSS and uniform VMSS instances are different. Procedure I fixed this in P #1451. In addition, if you want to obtain uniform vmss instances, set the parameter 'include_vmss_resource_groups' in your dynamic inventory scripts. Thank you!
```e.g.
/subscriptions/xxxxxxx/resourceGroups/v-xisuRG-uniform/providers/Microsoft.Compute/virtualMachineScaleSets/uniform-vmss/virtualMachines/0
/subscriptions/xxxxxxx/resourceGroups/v-xisuRG-flexible/providers/Microsoft.Compute/virtualMachines/flexible-vmss_6b8cb26f
```
sample as:
```e.g.
plugin: azure.azcollection.azure_rm
include_vmss_resource_groups:
    - ****-uniform
    - ****-flexible
******************
auth_source: auto

nskidotgo:
I am also running into this issue, we have both flexible and uniform scalesets in the same RG. Any workarounds? thanks

PR #1440, opened 2024-02-09
Added possibility to assign RG name via env var ANSIBLE_AZURE_VMSS_RESOURCE_GROUPS for VMSS and Fix inventory fail when trying to parse instance from the flexible scale set

### LINKED ISSUE #1432: Add possibility to assign VMSS rg name via environment variable
##### SUMMARY
Add the possibility to assign VMSS rg name **include_vmss_resource_groups** via an environment variable, for example, **ANSIBLE_AZURE_VMSS_RESOURCE_GROUPS**, like we can do, in the case of the **include_vm_resource_groups** with the environment variable **ANSIBLE_AZURE_VM_RESOURCE_GROUPS**

##### ISSUE TYPE
- Feature Idea

Justwmz:
##### SUMMARY

Added possibility to assign RG name via env var **ANSIBLE_AZURE_VMSS_RESOURCE_GROUPS** for VMSS
closes https://github.com/ansible-collections/azure/issues/1432

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,77 @@
 # Change Log
 
+## v3.0.0 (2024-10-31)
+
+### NEW MODULES
+  - azure_rm_virtualnetworkgatewayconnection: Support to managed virtual network gateway connection's resource ([#1627](https://github.com/ansible-collections/azure/pull/1627))
+  - azure_rm_virtualnetworkgatewayconnection_info: Support to fetch the virtual network gateway connection resource ([#1627](https://github.com/ansible-collections/azure/pull/1627))
+  - azure_rm_mysqlflexibleserver: Support managed flexible server ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexibleserver_info: Support to fetch managed flexible server ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexibleconfiguration: Support managed flexible server configuration ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexibleconfiguration_info: Support to fetch managed flexible server configuration ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexibledatabase: Support managed flexible server database ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexibledatabase_info: Support to fetch flexible server database ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexiblefirewallrule: Support managed flexible server firewall rule ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_rm_mysqlflexiblefirewallrule_info: Support to fetch flexible server firewall rule ([#1586](https://github.com/ansible-collections/azure/pull/1586))
+  - azure_identity_multiple_user: Managed identity cleanup ([#1724](https://github.com/ansible-collections/azure/pull/1724))
+  - azure_identity_single: Managed identity cleanup ([#1724](https://github.com/ansible-collections/azure/pull/1724))
+  - azure_rm_imagesku_info: VM Image SKUs support in Ansible collection for Azure module creation ([#1719](https://github.com/ansible-collections/azure/pull/1719))
+
+### FEATURE ENHANCEMENT
+  - azure_rm_eventhub: Add support for managed identity ([#1696](https://github.com/ansible-collections/azure/pull/1696))
+  - test case:
+    - Use the password plug-in to generate a login key ([#1699](https://github.com/ansible-collections/azure/pull/1699))
+    - Set the value of open port ([#1704](https://github.com/ansible-collections/azure/pull/1704))
+    - Update `azure_rm_sqlelasticpool` test case ([#1714](https://github.com/ansible-collections/azure/pull/1714))
+  - azure_rm_openshiftmanagedcluster: Add support `outbound_type` ([#1664](https://github.com/ansible-collections/azure/pull/1664))
+  - requirements.txt:
+    - Upgrade `azure-mgmt-containerinstance` to `v10.1.0` ([#1709](https://github.com/ansible-collections/azure/pull/1709))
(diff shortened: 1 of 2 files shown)


Justwmz:
##### SUMMARY

Fix inventory failure when trying to parse instances from the flexible scale set
closes https://github.com/ansible-collections/azure/issues/1431


##### ISSUE TYPE

- Bug Fix

##### COMPONENT NAME

azure_rm

plugins/inventory/azure_rm.py
```diff
@@ -403,6 +407,12 @@ def _on_vmss_page_response(self, response):
         # FUTURE: add direct VMSS filtering by tag here (performance optimization)?
         for vmss in response['value']:
             url = '{0}/virtualMachines'.format(vmss['id'])
+
+            # Since Flexible instance is a standalone VM, we are replacing ss provider to the vm one in order to allow inventory get instance view.
+            if vmss['properties']['orchestrationMode'] == 'Flexible':
+                newProvider = 'Microsoft.Compute/virtualMachineScaleSets/{0}/virtualMachines'.format(vmss['name'])
+                url = url.replace(newProvider, "Microsoft.Compute/virtualMachines")
```
Fred-sun: @Justwmz In this way, uniform VMSS instances can be obtained. However, because flexible VMSS instances are similar to regular VMS, the flexible VMSS instances will be obtained twice. Is there any way to avoid this? Thank you!

Justwmz: @Fred-sun Yes, there is the way to filter them out by the properties object, each flexible instance has a property which is indicated to which flexible scale set it is related.
I am just right now trying to test it, to see if that will works.

Fred-sun: @Justwmz Thanks for your feedback! Look forward a good results! Thank you very much!

Justwmz: @Fred-sun I tried multiple options to get it to work properly.
If you have VM and VMss RG's assigned to the same name where both types of scale sets are sitting, the call for pulling all VMs within RG will receive a list of those instances, since they are standalone VMs.

Maybe the better option will be to just obtain them as standalone VMs since they are pure VMs and need to be managed separately, not like in the Uniform instances which are managed by their parent?

P.S. I get one thing that works, which will parse the name of the instances from the Flexible scale set and then send a request per instance to obtain VM details. (But anyway, I still think that the ideal solution will be is ignore flexible parent since instances are managed separatly and just obtain them as VMs)

Justwmz: @Fred-sun  To ignore flexible parent we just need to change this:
```
        for vmss in response['value']:
            url = '{0}/virtualMachines'.format(vmss['id'])

            # Since Flexible instance is a standalone VM, we are replacing ss provider to the vm one in order to allow inventory get instance view.
            if vmss['properties']['orchestrationMode'] == 'Flexible':
                newProvider = 'Microsoft.Compute/virtualMachineScaleSets/{0}/virtualMachines'.format(vmss['name'])
                url = url.replace(newProvider, "Microsoft.Compute/virtualMachines")

            # VMSS instances look close enough to regular VMs that we can share the handler impl...
            self._enqueue_get(url=url, api_version=self._compute_api_version, handler=self._on_vm_page_response, handler_args=dict(vmss=vmss))
```
To this:
```
        for vmss in response['value']:
            url = '{0}/virtualMachines'.format(vmss['id'])

            # Since Flexible instance is a standalone VM, we are ignoring parsing instances data from parent and just obtaining them as regular VMs.
            if vmss['properties']['orchestrationMode'] != 'Flexible':
            # VMSS instances look close enough to regular VMs that we can share the handler impl...
               self._enqueue_get(url=url, api_version=self._compute_api_version, handler=self._on_vm_page_response, handler_args=dict(vmss=vmss))

```
And it will work fine in all cases.
Maybe in the future MS will add support to get flexible ss instances in the same way as for the Uniform.

Fred-sun: @Justwmz That sounds great, but in my tests I found that I was returning regular VM information, not VMSS information! Thank you!
```e.g.
Return info:
vmss: {}

Actual VMSS info：
                "vmss": {
                    "id": "/subscriptions/xxx/resourceGroups/v-xisuRG-uniform/providers/Microsoft.Compute/virtualMachineScaleSets/flexible-vmss-name",
                    "name": "flexible-vmss"
                }

Justwmz: @Fred-sun Yes you will be able to get info about scale set itself and list of the instances. But the key difference is that when you are obtaining a list of the instances in uniform scale set, it give you a full instance information.
In case of flexible scale set it will only present a limited information, like location, azure zone and 2-3 more parameters.

In order to get a full info about instance we need to use same api call as we are using for the standalone vms.

That’s why I basically propose ignore parent of the flexible scale set completely and only relay on the standalone vm api, since instances in the flexible scale set is standalone vms.

Fred-sun: @Justwmz Yes, I agree with the idea of returning only VM or only the flexible VMSS instance VMS. VM are all virtual machines with a large range, while flexible VMSS instances are more precise. So we can filter and return flexible VMSS instance information. It's more accurate. Thank you!

Fred-sun:
@Justwmz According to the return range, should we return VMSS instances and filter VMSS instances in VMS? right?

Justwmz:
@Fred-sun We can do that, but there will be another problem.
If you supply the name of the rg for the include_vm_resource_groups and include_vmss_resource_groups to the same value.
It will cause a double fetch of instances within a scale set because they are regular VMS. After all, they are a part of both types...

We can add a check when we fetch vms if it is a part of the scale set, but you need to perform an API call to Azure to see from the response if this property is present or not.
Which will anyway cause a double API call.

So in our case, better to completely filter out flexible parents, and let inventory fetch flexible instances using --> include_vm_resource_groups.

This solves a problem with double API calls for the same instances and from the performance side it will be better.

Fred-sun:
@Justwmz Your idea is really acceptable, but we still need to discuss it, thank you!

Justwmz:
Hi @Fred-sun,
Is there any news regarding this issue? Any decisions being made?

Fred-sun:
@Justwmz There is a conflict, please fix it. Thank you!

Issue #1438, opened 2024-02-08
azure_rm: Dynamic Inventory allow for template expressions in some parameters


franciscogarcia-payroc:
##### SUMMARY
We would love to be able to use templates to evaluate the values of client_id, secret, tenant and subscription_id when configuring authentication for the azure_rm inventory plugin. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`azure_rm` inventory plugin

##### ADDITIONAL INFORMATION
 Our use case is somewhat particular, and we can't leverage the cli or environment variables for various reasons and we want to avoid our secrets hard coded in file. The ideal way for us to get those values in without having them on a file on disk is to execute a lookup in a template. This behavior works as described in a few other inventory plugins we're using for enumerating VMs on other providers, but in this case the plugin only seems to accept static values in the following fields:

`client_id`
`secret`
`tenant`
`subscription_id`

To prove the case and get us past this issue in our current proof of concept we added the following snippet to [azure_rm.py](https://github.com/ansible-collections/azure/blob/dev/plugins/inventory/azure_rm.py#L252) in the _credential_setup method. 

This seemed to work as intended to allow for the lookup inside the inventory plugin to be executed dynamically.

```python
  def _credential_setup(self):
          auth_options = dict(
              auth_source=self.get_option('auth_source'),
              profile=self.get_option('profile'),
              subscription_id=self.get_option('subscription_id'),
              client_id=self.get_option('client_id'),
              secret=self.get_option('secret'),
              tenant=self.get_option('tenant'),
              ad_user=self.get_option('ad_user'),
              password=self.get_option('password'),
              cloud_environment=self.get_option('cloud_environment'),
              cert_validation_mode=self.get_option('cert_validation_mode'),
              api_profile=self.get_option('api_profile'),
              track1_cred=True,
              adfs_authority_url=self.get_option('adfs_authority_url')
          )

       #this part here was added
       if self.templar.is_template(auth_options["tenant"]):
            auth_options["tenant"] = self.templar.template(variable=auth_options["tenant"], disable_lookups=False)

        if self.templar.is_template(auth_options["client_id"]):
            auth_options["client_id"] = self.templar.template(variable=auth_options["client_id"], disable_lookups=False)

        if self.templar.is_template(auth_options["secret"]):
            auth_options["secret"] = self.templar.template(variable=auth_options["secret"], disable_lookups=False)

        if self.templar.is_template(auth_options["subscription_id"]):
[...]

PR #1443, opened 2024-02-13
Allow for template expressions in some parameters

franciscogarcia-payroc:
##### SUMMARY
We would love to be able to use templates to evaluate the values of client_id, secret, tenant and subscription_id when configuring authentication for the azure_rm inventory plugin.

PR to address issue #1438 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
`azure_rm` inventory plugin

##### ADDITIONAL INFORMATION
Our use case is somewhat particular, and we can't leverage the cli or environment variables for various reasons and we want to avoid our secrets hard coded in file. The ideal way for us to get those values in without having them on a file on disk is to execute a lookup in a template. This behavior works as described in a few other inventory plugins we're using for enumerating VMs on other providers, but in this case the plugin only seems to accept static values in the following fields:

`client_id`
`secret`
`tenant`
`subscription_id`

the change would evaluate if the parameters passed are a template expression and would then resolve the expression before moving on to the `AzureRMAuth` method

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -250,6 +250,18 @@ class InventoryModule(BaseInventoryPlugin, Constructable):
             adfs_authority_url=self.get_option('adfs_authority_url')
         )
 
+        if self.templar.is_template(auth_options["tenant"]):
+            auth_options["tenant"] = self.templar.template(variable=auth_options["tenant"], disable_lookups=False)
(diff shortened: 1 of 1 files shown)


franciscogarcia-payroc:
closing after rename of branch, will reopen as `franciscogarcia-payroc/dev`

PR #1507, opened 2024-03-21
allow service principals and nested groups to be returned in membership attributes

### LINKED ISSUE #1506: allow service_principals and groups to appear in group_members output
##### SUMMARY
Currently, the modules for AD group manipulation only return membership using the `transitive_members` property, which collapses nested groups. Additionally, due to a known issue in the graph API, service principals are not returned in queries for any kind of group member. The only way to view service principals is by querying the group object and expanding the `members` property explicitly.

We should add a flag that forces group_membership return properties to be sourced from the expanded `members` property, which will properly expose nested groups and service principals. This flag will default to false to preserve compatibility with the existing roles.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_adgroup_info
azure_rm_adgroup

##### ADDITIONAL INFORMATION
See example invocation below

```yaml
  azure.azcollection.azure_rm_adgroup_info:
    object_id: "xx-yy-zz"
   return_group_members: true
   raw_membership: true # <-- new flag that specifies the returned members should include service principals and not be transitive
```

kent007:
##### SUMMARY

Fixes https://github.com/ansible-collections/azure/issues/1506

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_adgroup
azure_rm_adgroup_info

##### ADDITIONAL INFORMATION
I also fixed a few typos in docs and function names. Verified with some manual testing and a clean pass of ansible sanity tests

### DIFF
--- a/plugins/modules/azure_rm_adgroup.py
+++ b/plugins/modules/azure_rm_adgroup.py
@@ -63,6 +63,12 @@ options:
             - The azure ad objects asserted to not be owners of the group.
         type: list
         elements: str
+    raw_membership:
+        description:
+            - By default the group_members return property is flattened and partially filtered of non-User objects
+              before return. This argument disables those transformations.
+        default: false
+        type: bool
 extends_documentation_fragment:
     - azure.azcollection.azure
 author:
@@ -104,6 +110,15 @@ EXAMPLES = '''
       - "{{ ad_object_1_object_id }}"
       - "{{ ad_object_2_object_id }}"
 
+- name: Ensure Users are Members of a Group using object_id. Specify the group_membership return should be unfiltered
+  azure_rm_adgroup:
+    object_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
+    state: 'present'
+    present_members:
+      - "{{ ad_object_1_object_id }}"
+      - "{{ ad_object_2_object_id }}"
+    raw_membership: true
+
 - name: Ensure Users are not Members of a Group using display_name and mail_nickname
   azure_rm_adgroup:
     display_name: "Group-Name"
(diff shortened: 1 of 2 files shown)


plugins/modules/azure_rm_adgroup_info.py
```diff
@@ -84,6 +89,13 @@
     object_id: xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx
     return_owners: true
     return_group_members: true
+
```
Fred-sun: remove trailing space!
```suggestion

```

plugins/modules/azure_rm_adgroup_info.py
```diff
@@ -55,6 +55,11 @@
             - Indicate whether the groups in which a groups is a member should be returned with the returned groups.
         default: False
         type: bool
+    raw_membership:
+        description:
+            - By default the group_members return property is flattened and partially filtered of non-User objects before return. This argument disables those transformations.
```
Fred-sun: line too long (175 > 160 characters)
```suggestion
            - By default the group_members return property is flattened and partially filtered of non-User objects before return.\
              This argument disables those transformations.
```

plugins/modules/azure_rm_adgroup.py
```diff
@@ -63,6 +63,11 @@
             - The azure ad objects asserted to not be owners of the group.
         type: list
         elements: str
+    raw_membership:
+        description:
+            - By default the group_members return property is flattened and partially filtered of non-User objects before return. This argument disables those transformations.
```
Fred-sun:
```suggestion
            - By default the group_members return property is flattened and partially filtered of non-User objects before return. \
             This argument disables those transformations.
```

xuzhang3 (approved):

Issue #1041, opened 2022-12-09
Feature Request - Add support for Azure Front Door


paulczar:
Add support from Azure Front Door.

see https://learn.microsoft.com/en-us/cli/azure/service-page/azure%20front%20door?view=azure-cli-latest for the various CLI commands we should make available.

jartoo:
This would be a good addition! +1 for feature request..

jartoo:
Ok, for clarity here, the new version of the "Front Door Service" is actually managed out of the azure-mgmt-cdn library using the CdnManagementClient class, which is counter to what one would expect from the naming conventions.  Currently, you offer the azure.azcollection.azure_rm_cdnprofile module, which is missing support for the Standard_AzureFrontDoor and Premium_AzureFrontDoor options.  As well, additional modules for the Origin Group, Origin, Rule Set, Rules, End Points, etc.. would be needed.  Not sure if it makes sense to add modules for each of these objects, or only some of them with the rest carried in the cdnprofile module.  Interested in thoughts on this...

jartoo:
Here is the link to the Python SDK relevant for the Standard and Premium Front Door types:

https://learn.microsoft.com/en-us/python/api/azure-mgmt-cdn/azure.mgmt.cdn.cdnmanagementclient?view=azure-python

Objects which will need Ansible Module Support as part of the Front Door Service:
Profiles
Custom Domains
End Points
Origin Groups
Origins
Routes
Rule Sets
Rules
Security Policies
Secrets
Log Analytics

..though not sure if this is exhaustive

PR #1604, opened 2024-06-18
Add new module azure_rm_afdrules to support Azure Frontdoor Standard and Premium

jartoo:
##### SUMMARY

This module manages Frontdoor Rules for the Standard and Premium services. I propose building this as a separate module in lieu of complicating the azure_cdn_profile module so as to make the management of each of the elements easier. Azure Frontdoor is quite complex, so matching the modules with the Python SDK APIs seems to be the prudent design choice.

Continues to work on https://github.com/ansible-collections/azure/issues/1041 This does not complete this issue, yet.

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

Module: azure_rm_afdrules

##### ADDITIONAL INFORMATION

Provides the ability to manage Rules, per the Python SDK here: https://learn.microsoft.com/en-us/python/api/azure-mgmt-cdn/azure.mgmt.cdn.operations.rulesoperations?view=azure-python

Integration Tests: passed
Sanity Tests: passed

```paste below

```

### DIFF
--- a/meta/runtime.yml
+++ b/meta/runtime.yml
@@ -14,6 +14,8 @@ action_groups:
     - azure.azcollection.azure_rm_adserviceprincipal_info
     - azure.azcollection.azure_rm_aduser
     - azure.azcollection.azure_rm_aduser_info
+    - azure.azcollection.azure_rm_afdrules
+    - azure.azcollection.azure_rm_afdrules_info
     - azure.azcollection.azure_rm_aks
     - azure.azcollection.azure_rm_aks_info
     - azure.azcollection.azure_rm_aksagentpool
--- a/plugins/modules/azure_rm_afdrules.py
+++ b/plugins/modules/azure_rm_afdrules.py
@@ -1,0 +1,1144 @@
+#!/usr/bin/python
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+#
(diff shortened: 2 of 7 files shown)


Fred-sun:
@jartoo self.actions and conditions,  if not configured in the playbook, this parameter is None, so before comparing, please make sure that this value is not None, thank you!

Fred-sun:
@jartoo plugins/module_utils/azure_rm_common_ext.py has a function 'default_compare' for comparison, For example, self.default_compare({}, new_parameters, old_parameters, '', dict(compare=[])), You can take a try！ Thank you very much!

jartoo:
Will take a look at this...thanks!

jartoo:
Using idempotent_check() would work if the structures are identical.  The problem with using that function in this case is that each Action and Condition type_name changes the variables that are used.  I also do not force the user of the module to declare the type_name for each and hard code that in based on the name of the Action and Condition to simplify usage.  Because the structures of the dict is therefore different, I am not sure I can use that function as the query from the Python SDK returns a structure that is actually a subset of all the possible parameters.  But this did open my eyes to the AzureRMModuleBaseExt class and some of the options you have built in there.  Please correct me if I am misunderstanding the use case applicability here.  I will work on checking for None however, and resubmit with those changes shortly.  Thanks!

jartoo:
The structure is also slightly different, where I flattened the parameters section to reduce the hierarchy complexity...

Fred-sun:
@jartoo You seem to have restored the changes of the dev branch, please check this PR again, thank you!

Fred-sun:
Move to #1885, thanks!

PR #1621, opened 2024-07-03
Add Managed Identity support to CdnProfile module

nirarg:
Changed module: azure_rm_cdnprofile

##### SUMMARY
Add Managed Identity support to CdnProfile module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
azure_rm_cdnprofile

### DIFF
--- a/plugins/modules/azure_rm_cdnprofile.py
+++ b/plugins/modules/azure_rm_cdnprofile.py
@@ -56,6 +56,7 @@ options:
 extends_documentation_fragment:
     - azure.azcollection.azure
     - azure.azcollection.azure_tags
+    - azure.azcollection.azure_identity_multiple
 
 author:
     - Hai Cao (@caohai)
@@ -85,11 +86,11 @@ id:
     example:
             id: /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourcegroups/myResourceGroup/providers/Microsoft.Cdn/profiles/myCDN
 '''
-from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common import AzureRMModuleBase
+from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common_ext import AzureRMModuleBaseExt
 import uuid
 
 try:
-    from azure.mgmt.cdn.models import Profile, Sku
+    from azure.mgmt.cdn.models import Profile, Sku, ManagedServiceIdentity, UserAssignedIdentity, ProfileUpdateParameters
     from azure.mgmt.cdn import CdnManagementClient
 except ImportError as ec:
     # This is handled in azure_rm_common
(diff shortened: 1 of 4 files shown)


nirarg:
There is still an issue with using `append=false` in the user identity
The module works as expected but its seems that the API still performing append instead of overwrite
I'm investigating this issue

nirarg:
Hi @Fred-sun,

Can you please advise on an issue I'm encountering in this PR?

When I update the Identity with "append=false", I expect the user identity to be replaced with the new one, not appended. However, the new task I added to the integration test in this PR (task name: "Update the CDN profile - replace user identity") doesn't return the expected result.

While debugging this, I couldn't find anything suspicious that could be causing this issue. I would appreciate any help you can provide.

Thank you!

plugins/modules/azure_rm_cdnprofile.py
```diff
@@ -177,6 +194,11 @@ def exec_module(self, **kwargs):
 
         if self.state == 'present':
 
+            curr_identity = response["identity"] if response else None
+            update_identity = False
+            if self.identity:
+                update_identity, self.identity = self.update_managed_identity(curr_identity=curr_identity, new_identity=self.identity)
```
Fred-sun: @nirarg How did you set 'append=False'? The 'allow_identities_append=True' parameter of the function self.update_managed_identity has a default value, so it is added. If you want to override, you need to pass the function when 'allow_identities_append=False'. So you should pass it to the function according to the arguments of 'append'. Thank you!

nirarg: Hi @Fred-sun,
Yes, I used `append=false`. The `allow_identities_append` variable is used for cases where the `append` attribute is not available. This is not the case here, `append` is being used. Therefore, the default value in the function is `allow_identities_append=True`.
I debugged the code and saw that the `identity` is returned as expected from this function and the result is sent to the API. Therefore, I suspect there is something I miss in the API call or in the SDK code.

p3ck:
I have this passing with the included integration tests when using the fixes I pushed in #1635 .  You just need to update this pr with the following:

```
diff --git a/plugins/modules/azure_rm_cdnprofile.py b/plugins/modules/azure_rm_cdnprofile.py
index 0b371d2..eeec910 100644
--- a/plugins/modules/azure_rm_cdnprofile.py
+++ b/plugins/modules/azure_rm_cdnprofile.py
@@ -197,7 +197,7 @@ class AzureRMCdnprofile(AzureRMModuleBaseExt):
             curr_identity = response["identity"] if response else None
             update_identity = False
             if self.identity:
-                update_identity, self.identity = self.update_managed_identity(curr_identity=curr_identity, new_identity=self.identity)
+                update_identity, self.identity = self.update_managed_identity(curr_identity=curr_identity, new_identity=self.identity, patch_support=True)
 
             if not response:
                 self.log("Need to create the CDN profile")
```

The reason seems to be that some of the API calls (update) use the PATCH method.  The backend logic will append to the list unless you pass keys with a value of NONE, then they will be removed.

nirarg:
The verification CI test was started here:
https://dev.azure.com/azclitools/public/_build/results?buildId=174991&view=results

nirarg:
Hi @Fred-sun @xuzhang3 
Is there any update regarding this PR review?
Is there anything else required to complete here before merge?

xuzhang3 (approved):

PR #1625, opened 2024-07-09
Wait 10 minutes to get the imageversion state

Fred-sun:
##### SUMMARY

Wait 10 minutes to get the imageversion state. If it is always' creating', an error message is displayed"， similar with #1584

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm_galleryimageversion.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_galleryimageversion.py
+++ b/plugins/modules/azure_rm_galleryimageversion.py
@@ -855,9 +855,13 @@ class AzureRMGalleryImageVersions(AzureRMModuleBaseExt):
         else:
             self.fail("Create or Updating fail, no match message return, return info as {0}".format(response))
 
+        i = 0
         while response['properties']['provisioningState'] == 'Creating':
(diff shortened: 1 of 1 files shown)


jeremycline:
FWIW this doesn't fix what #1584 addressed. The problem is that `get_resource()` returns both a boolean and a valid response object, but the caller doesn't handle that. The boolean can be returned when a transient network error occurs, or when (I assume) there's a race and the resource doesn't exist quite yet.

Adding a retry limit is good and I agree that it should not try forever, but I think it's worth noting this is a different issue than #1584.

Fred-sun:
@jeremycline However, if you re-obtain the resource after the resource is updated, if the resource cannot be obtained, return the previous status of the resource, which is not appropriate, and will give the user an error message. Thank you!

xuzhang3 (approved):

PR #1626, opened 2024-07-09
Update identity support in sqlserver

p3ck:
##### SUMMARY
Update identity to support UserAssigned and
SystemAssigned,UserAssigned.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_sqlserver.py
plugins/modules/azure_rm_sqlserver_info.py
tests/integration/targets/azure_rm_sqlserver/tasks/main.yml

##### ADDITIONAL INFORMATION
This expands the managed identity support from just SystemAssigned to include UserAssigned as well.

### DIFF
--- a/plugins/modules/azure_rm_sqlserver.py
+++ b/plugins/modules/azure_rm_sqlserver.py
@@ -338,7 +338,7 @@ class AzureRMSqlServer(AzureRMModuleBaseExt):
             if key not in valid_user_assigned_keys:
                 errors.append("Invalid key {0}".format(key))
             if key == "append":
-                if isinstance(key, bool) is not True:
+                if isinstance(self.identity['user_assigned_identities'][key], bool) is not True:
                     errors.append("identity->user_assigned_identity->append must be True or False")
 
         if errors:
--- a/plugins/modules/azure_rm_sqlserver.py
+++ b/plugins/modules/azure_rm_sqlserver.py
@@ -45,9 +45,26 @@ options:
         type: str
     identity:
         description:
-            - The identity type. Set this to C(SystemAssigned) in order to automatically create and assign an Azure Active Directory principal for the resource.
-            - Possible values include C(SystemAssigned).
+            - Specifies the managed identity to be used with SqlServer.
+            - If a string, the Managed identity type is based on the name.
+            - Possible values include C(SystemAssigned) and C(None).
+            - If a dict with the keys I(type), I(user_assigned_identities)
+            - Possible values for I(type) include C(SystemAssigned, C(UserAssigned),
+              C(SystemAssigned, UserAssigned), and C(None).
+            - Possible values for I(user_assigned_identities) include a dict
+              with I(id) and I(append).
+            - Possible values for I(id) is a list of user assigned identities ID's
+            - Possible values for I(append) is a boolean of True to append identities
+              or False to overwrite with new I(id)'s.
+            - The string format is (deprecated) and the new dict format should be used
+              going forward.
+        type: raw
+    primary_user_assigned_identity_id:
+        description:
+            - Specifies the primary User Assigned Identity to use.
+            - This is required if you are using Managed Identity type of UserAssigned
         type: str
+        version_added: "2.7.0"
     minimal_tls_version:
         description:
             - Require clients to use a specified TLS version.
@@ -197,6 +214,7 @@ from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common
 try:
     from azure.core.exceptions import ResourceNotFoundError
     from azure.core.polling import LROPoller
+    from azure.mgmt.sql.models import (ResourceIdentity, UserIdentity)
 except ImportError:
     # This is handled in azure_rm_common
     pass
@@ -243,7 +261,10 @@ class AzureRMSqlServer(AzureRMModuleBaseExt):
                 type='str'
             ),
             identity=dict(
-                type='str'
+                type='raw'
+            ),
+            primary_user_assigned_identity_id=dict(
+                type='str',
             ),
             minimal_tls_version=dict(
                 type="str",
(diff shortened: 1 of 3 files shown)


tests/integration/targets/azure_rm_sqlserver/tasks/main.yml
```diff
@@ -37,13 +64,27 @@
       - output.changed
       - output.state == 'Ready'
 
+- name: Gather facts SQL Server
+  azure_rm_sqlserver_info:
+    resource_group: "{{ resource_group }}"
+    server_name: "sqlsrv{{ random_postfix }}"
+  register: output
+- name: Debug
+  debug:
```
Fred-sun: Use `ansible.builtin.debug` or `ansible.legacy.debug` instead
```suggestion
  ansible.builtin.debug:
```

tests/integration/targets/azure_rm_sqlserver/tasks/main.yml
```diff
@@ -70,6 +117,20 @@
       - output.changed
       - output.state == 'Ready'
 
+- name: Gather facts SQL Server
+  azure_rm_sqlserver_info:
+    resource_group: "{{ resource_group }}"
+    server_name: "sqlsrv-extended-{{ random_postfix }}"
+  register: output
+- name: Debug
+  debug:
```
Fred-sun: Use ansible.builtin.debug or ansible.legacy.debug instead
```suggestion
  ansible.builtin.debug:
```

Fred-sun:
@p3ck small changed request!

p3ck:
> @p3ck small changed request!

All set.  I meant to remove the debug task before submitting it.  Thanks!

plugins/modules/azure_rm_sqlserver.py
```diff
@@ -277,16 +298,55 @@ def __init__(self):
...
+            errors.append("Invalid identity->type, Valid choices are: [{0}]".format(type_choices))
+        for key in self.identity.get("user_assigned_identities", dict()).keys():
+            if key not in valid_user_assigned_keys:
+                errors.append("Invalid key {0}".format(key))
+            if key == "append":
+                if isinstance(key, bool) is not True:
+                    errors.append("identity->user_assigned_identity->append must be True or False")
+                # Default option for append is True
+                self.identity["user_assigned_identities"]["append"] = \
+                    self.identity["user_assigned_identities"].get("append", True)
```
nirarg: Why do you need this assignment here?
This is done inside the if statement `key == "append"` is true, which I think is wrong
Need to assign the default value in case that `append` key doesn't exist

plugins/modules/azure_rm_sqlserver.py
```diff
@@ -243,7 +261,10 @@ def __init__(self):
                 type='str'
             ),
             identity=dict(
-                type='str'
+                type='raw'
+            ),
+            primary_user_assigned_identity_id=dict(
```
nirarg, line 266: Where is it used?
I see only the assignment but not the usage

plugins/modules/azure_rm_sqlserver.py
```diff
@@ -313,6 +373,17 @@ def exec_module(self, **kwargs):
 
         old_response = self.get_sqlserver()
 
+        if self.identity and isinstance(self.identity, dict):
+            self.validate_identity_parameter()
+            update_identity, identity = self.update_managed_identity(old_response and old_response.get('identity'),
+                                                                     self.identity)
+            if update_identity:
+                self.parameters.update({"identity": identity.as_dict()})
+        elif self.identity and isinstance(self.identity, str):
```
nirarg, line 379: Please add here a warning message for user that this is deprecated and `identity` type need to be used here

plugins/modules/azure_rm_sqlserver.py
```diff
@@ -277,16 +298,55 @@ def __init__(self):
...
 
+    @property
+    def managed_identity(self):
+        if not self._managed_identity:
+            self._managed_identity = {"identity": ResourceIdentity,
+                                      "user_assigned": UserIdentity
+                                      }
+        return self._managed_identity
+
+    def validate_identity_parameter(self):
```
nirarg, line 322:
```suggestion
    def assert_identity_params(self):
```
This function fails in case that the parameters are wrong, so maybe this name is more appropriate
Just a suggestion, Up to you

plugins/modules/azure_rm_sqlserver.py
p3ck: Good suggestion,  Thanks.

TASK [azure_rm_sqlserver : Create instance of SQL Server -- check mode] ********
[WARNING]: Specifying the identity type as a string is deprecated, please
update your playbook.
changed: [testhost]

plugins/modules/azure_rm_sqlserver.py
p3ck: It's done here.  Only needs to be added to parameters
```
               elif key == "primary_user_assigned_identity_id":
                    self.parameters.update({"primary_user_assigned_identity_id": kwargs[key]})
```

plugins/modules/azure_rm_sqlserver.py
p3ck: I think using assert in the name would imply that it would be optimized out with -O.  From my understanding of best security practices we should not use assert statements in general code, only in test cases.

plugins/modules/azure_rm_sqlserver.py
p3ck: good point.  and the update_managed_identity handles the default.

plugins/modules/azure_rm_sqlserver.py
nirarg: Oh, I got it
So this is sent to the API as is without any manipulation or validation in the ansible code, right?

nirarg (dismissed):

plugins/modules/azure_rm_sqlserver.py
```diff
@@ -277,16 +298,52 @@ def __init__(self):
...
+                errors.append("Invalid key {0}".format(key))
+        # Default option for type is "None"
+        self.identity["type"] = self.identity.get("type", "None")
+        if self.identity.get("type") not in type_choices:
+            errors.append("Invalid identity->type, Valid choices are: [{0}]".format(type_choices))
+        for key in self.identity.get("user_assigned_identities", dict()).keys():
+            if key not in valid_user_assigned_keys:
+                errors.append("Invalid key {0}".format(key))
+            if key == "append":
+                if isinstance(key, bool) is not True:
```
Fred-sun:
```suggestion
                if isinstance(self.identity['user_assigned_identities'][key], bool) is not True:
```

Fred-sun (commented):
@p3ck Small changes required!

tests/integration/targets/azure_rm_sqlserver/tasks/main.yml
p3ck: Thank you.  I committed your suggestion

p3ck:
> @p3ck Small changes required!

Thanks!  I committed your suggestion

tests/integration/targets/azure_rm_sqlserver/tasks/main.yml
p3ck: I removed the debug tasks

tests/integration/targets/azure_rm_sqlserver/tasks/main.yml
p3ck: I removed the debug tasks

p3ck:
Ping - Can you look at this again?  Thanks

Fred-sun (approved):

xuzhang3 (approved):

PR #1635, opened 2024-07-16
Fix update_managed_identity to handle removing entries

p3ck:
##### SUMMARY
Some API endpoints will append UserAssignedIdentities without removing the original ones.  To remove the original entries you must specify None for the value of the key you want to remove.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/azure_rm_common_ext.py

##### ADDITIONAL INFORMATION

We need to add append=False test cases to the existing modules using this method.

### DIFF
--- a/plugins/module_utils/azure_rm_common_ext.py
+++ b/plugins/module_utils/azure_rm_common_ext.py
@@ -257,7 +257,7 @@ class AzureRMModuleBaseExt(AzureRMModuleBase):
             else:
                 return True
 
-    def update_single_managed_identity(self, curr_identity, new_identity):
+    def update_single_managed_identity(self, curr_identity, new_identity, patch_support=False):
         # Converting from single_managed_identity_spec to managed_identity_spec
         new_identity = new_identity or dict()
         new_identity_converted = {
@@ -268,9 +268,12 @@ class AzureRMModuleBaseExt(AzureRMModuleBase):
             new_identity_converted['user_assigned_identities'] = {
                 'id': [user_assigned_identity]
             }
-        return self.update_managed_identity(curr_identity=curr_identity, new_identity=new_identity_converted, allow_identities_append=False)
+        return self.update_managed_identity(curr_identity=curr_identity,
+                                            new_identity=new_identity_converted,
+                                            allow_identities_append=False, patch_support=patch_support)
 
-    def update_managed_identity(self, curr_identity=None, new_identity=None, allow_identities_append=True):
+    def update_managed_identity(self, curr_identity=None, new_identity=None,
+                                allow_identities_append=True, patch_support=False):
         curr_identity = curr_identity or dict()
         # TODO need to remove self.module.params.get('identity', {})
         # after changing all modules to provide the "new_identity" parameter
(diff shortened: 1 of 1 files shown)


p3ck:
Looks like the behavior depends on how the client is making the REST call.  While testing this change with the iot hub module I get the following error:

```
The 'UserAssignedIdentities' property keys should only be empty null during PATCH requests."
```

So it looks like we will need to add an option to the method depending on how the client code is making the call.  I propose we add a keyword called patch_support=False by default and enable this only on these modules that use the PATCH call.

nirarg (dismissed):
This is look good
@xuzhang3 @Fred-sun can you please also take a look?
We need this to be merged in order to unblock other PRs

xuzhang3:
@p3ck  test failed.   The API version(`2023-06-30-preview`) is not available in the iothub SDK
```log
The full traceback is:
Traceback (most recent call last):
  File "/tmp/ansible_azure_rm_iothub_payload_4pusx33w/ansible_azure_rm_iothub_payload.zip/ansible_collections/azure/azcollection/plugins/modules/azure_rm_iothub.py", line 908, in create_or_update_hub
  File "/home/eric/workspace/ansible-workspace/mymodule/test/lib/python3.10/site-packages/azure/mgmt/iothub/_iot_hub_client.py", line 269, in iot_hub_resource
    raise ValueError("API version {} does not have operation group 'iot_hub_resource'".format(api_version))
ValueError: API version 2023-06-30-preview does not have operation group 'iot_hub_resource'

During handling of the above exception, another exception occurred:

```

xuzhang3:
The iothub tests is disable in the CI pipe so I tested manually.

p3ck:
I removed the iothub testcase updates and will do those updates in a separate PR.

xuzhang3:
@p3ck  cannot replace the  identities. For example: Current identities is `test1` and `test2`,  I want to remove `test2` so I change the identity to `test1` but this does not seem to work.  The identities in the resource will still be `test1` and `test2`

p3ck:
> @p3ck cannot replace the identities. For example: Current identities is `test1` and `test2`, I want to remove `test2` so I change the identity to `test1` but this does not seem to work. The identities in the resource will still be `test1` and `test2`

How are you testing this?  Depending on the client and how the update is implemented you may need to enable patch_support=True.

If the client has begin_create_or_update it most likely is using POST restAPI and will work without this patch.  The full list of identities will be replaced with what we send.

If the client has begin_update it most likely is using PATCH restAPI and will need patch_support=True in order to remove the already existing entry. Example:

Object foo has identity test1 and test2.  We want to remove test2 so with PATCH we will need to send the following:

```
user_assigned_identities = { 'test1': {}, 'test2': None}
```

The None entry with test2 tells the server side to drop test2 identity.

But this only works for PATCH restAPI.  If using POST you simply send just 'test1': {} and omit 'test2'.  Which is why we need to flag in the method.

I hope this helps explain everything.

xuzhang3:
#1639 has been merged,   this PR can be closed

PR #1643, opened 2024-07-22
Add Managed Identity support to servicebus

p3ck:
##### SUMMARY
Adds Identity support to ServiceBus module

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_servicebus.py
plugins/modules/azure_rm_servicebus_info.py
tests/integration/targets/azure_rm_servicebus/tasks/main.yml

### DIFF
--- a/plugins/modules/azure_rm_servicebus.py
+++ b/plugins/modules/azure_rm_servicebus.py
@@ -47,6 +47,42 @@ options:
             - basic
             - premium
         default: standard
+    identity:
+        description:
+            - Identity for this resource.
+        type: dict
+        version_added: '2.7.0'
+        suboptions:
+            type:
+                description:
+                    - Type of the managed identity
+                choices:
+                    - SystemAssigned
+                    - SystemAssigned, UserAssigned
+                    - UserAssigned
+                    - 'None'
+                default: 'None'
+                type: str
+            user_assigned_identities:
+                description:
+                    - User Assigned Managed Identities and its options
+                required: false
+                type: dict
+                default: {}
+                suboptions:
+                    id:
+                        description:
+                            - List of the user assigned identities IDs associated to the WebApp
+                        required: false
+                        type: list
+                        elements: str
+                        default: []
+                    append:
+                        description:
+                            - If the list of identities has to be appended to current identities (true) or if it has to replace current identities (false)
+                        required: false
+                        type: bool
+                        default: True
 
 extends_documentation_fragment:
     - azure.azcollection.azure
@@ -75,7 +111,8 @@ id:
 '''
 
 try:
-    from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common import AzureRMModuleBase
+    from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common_ext import AzureRMModuleBaseExt
+    from azure.mgmt.servicebus.v2021_06_01_preview.models import (Identity, UserAssignedIdentity)
 except ImportError:
     # This is handled in azure_rm_common
     pass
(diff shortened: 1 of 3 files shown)


tests/integration/targets/azure_rm_servicebus/tasks/main.yml
```diff
@@ -23,13 +54,102 @@
...
+  azure_rm_servicebus:
+    name: "ns{{ rpfx }}"
+    resource_group: "{{ resource_group }}"
+    sku: premium
+    identity:
+      type: UserAssigned
+      user_assigned_identities:
+        id:
+          - "{{ user_identity_3 }}"
+        append: False
```
xuzhang3: Lint check failed
```log
WARNING  Unable to resolve FQCN for module azure_rm_recoveryservicesvault_info
WARNING  Listing 1 violation(s) that are fatal
yaml[truthy]: Truthy value should be one of [false, true]
tests/integration/targets/azure_rm_servicebus/tasks/main.yml:142

```

xuzhang3 (changes requested):

tests/integration/targets/azure_rm_servicebus/tasks/main.yml
p3ck: done - thanks

Fred-sun:
ready_for_review

plugins/modules/azure_rm_servicebus.py
```diff
@@ -47,6 +47,42 @@
...
+                        required: false
+                        type: list
+                        elements: str
+                        default: []
+                    append:
+                        description:
+                            - If the list of identities has to be appended to current identities (true) or if it has to replace current identities (false)
+                        required: false
+                        type: bool
+                        default: True
```
xuzhang3, line 85: Can we set `append` defaults to `false. The default behavior of the identities should be able added and removed not only added.

xuzhang3 (changes requested):

plugins/modules/azure_rm_servicebus.py
p3ck: I can change it but I was following the same default from azure_virtual_machine

https://github.com/ansible-collections/azure/blob/187b0c22ec3641d2b63db969509f0480352099f1/plugins/modules/azure_rm_virtualmachine.py#L492

which as been the default since 2023-06-15
```
ef944d9f (cpiment            2023-06-15 10:03:47 +0200  492)                         default: True
```

xuzhang3: My concern is that the default value of `append` is `true`, which means we can only add identities but not remove any. Users may be confused or they may simply treat it as a bug.

xuzhang3: Since we support `overwrite` and `append`, defaults to `overwrite`  would be better for users.

p3ck: I've changed it to default False.

We should put a PR through to change all of them to default False to be consistent.

plugins/modules/azure_rm_servicebus.py
```diff
@@ -47,6 +47,42 @@
...
+                        required: false
+                        type: list
+                        elements: str
+                        default: []
+                    append:
+                        description:
+                            - If the list of identities has to be appended to current identities (true) or if it has to replace current identities (false)
+                        required: false
+                        type: bool
+                        default: False
```
Fred-sun: @p3ck @p3ck The argument default value is ['true',](url https://github.com/ansible-collections/azure/blob/997e4de875fa12e99c9a064b614797a65c58677e/plugins/module_utils/azure_rm_common_ext.py#L37) Please keep consistent. Thank you!

p3ck: Ok - I updated to True.

We can create another PR to change the defaults to False for all the modules that use identity.

xuzhang3 (approved):

Issue #1614, opened 2024-06-27
Azure RM Inventory plugin does not return all hosts all the time


amitsharma-31:
##### SUMMARY
Syncing an Ansible inventory using the Azure RM Inventory plugin does not return all hosts all the time.
We sync our inventory on a schedule, but sometimes it barely returns any hosts.
For example last time it only captured 10 out of some 150+ hosts. Next run was again able to list all 150+ hosts

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
azure.azcollection.azure_rm (azure inventory plugin)

##### ANSIBLE VERSION

```
ansible [core 2.15.9]
  config file = ~/.ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/.venv/lib64/python3.9/site-packages/ansible
  ansible collection location = ~/.ansible/collections:/usr/share/ansible/collections
  executable location = ~/.venv/bin/ansible
  python version = 3.9.19 (main, May 16 2024, 08:45:40) [GCC 8.5.0 20210514 (Red Hat 8.5.0-22)] (~/.venv/bin/python3.9)
  jinja version = 3.1.2
  libyaml = True
```

##### COLLECTION VERSION

```
azure.azcollection             2.2.0
```

amitsharma-31:
It appears Ansible is making too many calls to Azure and is being throttled. Either Ansible should lower the number of calls or implement exponential backoff retry as described: https://learn.microsoft.com/en-us/azure/architecture/best-practices/retry-service-specific#general-rest-and-retry-guidelines. 

The changes are required to avoid throttling need to be configured.

Issue observed :- throttling (HTTP 429 errors) on the failed time slot. The throttling occurred on the ARM (Azure Resource Management) layer, not on the Resource Provider (Virtual Machines) layer. More details on throttling can be found at https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/request-limits-and-throttling

There is similar issue but way older https://github.com/ansible-collections/azure/issues/224

p3ck:
I'm trying to understand where it's failing.  I can see that the batch processing only handles 200 status code but I would expect it to fail in the batch request and it does seem to handle non 202 (it's a post call to /batch).

I think it would be helpful to add the following to the inventory plugin and see if we get any errors back.

```
diff --git a/plugins/inventory/azure_rm.py b/plugins/inventory/azure_rm.py
[...]

PR #1649, opened 2024-07-24
Implement long running polling

p3ck:
##### SUMMARY
This implements long running polling the same way the GenericRestClient does with a 600 second overall timeout and 30 second polling.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/inventory/azure_rm.py

##### ADDITIONAL INFORMATION
I ran the inventory integration tests but I don't have enough hosts in my azure instance to recreate the original issue.  So all I can do is verify that it still works with my limited test environment.

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -124,6 +124,7 @@ import json
 import re
 import uuid
 import os
+import time
 
 try:
     from queue import Queue, Empty
@@ -147,6 +148,7 @@ try:
     from azure.mgmt.core.tools import parse_resource_id
     from azure.core.pipeline import PipelineResponse
     from azure.mgmt.core.polling.arm_polling import ARMPolling
+    from azure.core.polling import LROPoller
 except ImportError:
     Configuration = object
     parse_resource_id = object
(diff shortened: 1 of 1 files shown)


p3ck:
Possible fix for #1614

Fred-sun:
@p3ck There's no problem on my end. But I don't have enough virtual machines on my side to repeat the problem. So I will push to merge #1614 users after they have no problem synchronizing updates. Thank you!

plugins/inventory/azure_rm.py
```diff
@@ -463,30 +472,61 @@ def _process_queue_batch(self):
                     # FUTURE: error-tolerant operation mode (eg, permissions)
                     # FUTURE: store/handle errors from individual handlers
                     result.handler(r['content'], **result.handler_args)
+                    batch_processed.append(returned_name)
+                elif status_code in _RETRY_CODES:
+                    # 429: Too many requests Error, Backoff and Retry
+                    batch_retry = True
+            if batch_retry:
+                time.sleep(backoff_factor * (2 ** (retry_count)))
```
xuzhang3, line 480: Regarding the backoff policy, retry time should based on the response. Other wise the retry request will be rejected too.
https://learn.microsoft.com/en-us/troubleshoot/azure/virtual-machines/windows/troubleshooting-throttling-errors#throttling-error-details

xuzhang3 (changes requested):

plugins/inventory/azure_rm.py
nirarg: Thank you @xuzhang3,
@p3ck is on vacation (he will be back next week) so I'm checking how to add the RetryPolicy class to be used here, in order to have this response retry time to be considered

nirarg: The issue here is that in the current implementation the wait sleep is done once for the all batch together
Do you have any suggestion/thoughts how to change it to work for each vm separately?

xuzhang3: @nirarg  I see, misunderstand the retry policy here.  The backoff policy and retry is focus on the local batch requests here so we cannot determine the retry time for all the request.

xuzhang3 (approved):

nirarg (approved):

plugins/inventory/azure_rm.py
Klaas-: https://github.com/Azure/azure-sdk-for-python/issues/36285 I think the retry stuff should be implemented on the python sdk level which azure_rm inventory uses. That could be a nice generic solution, but so far I have not convinced MSFT it's a good idea :)

PR #1651, opened 2024-07-25
Add Managed Identity support to Redis Cache

p3ck:
##### SUMMARY
Adds Managed Identity support to the rediscache module.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_rediscache.py
plugins/modules/azure_rm_rediscache_info.py
requirements.txt
tests/integration/targets/azure_rm_rediscache/tasks/main.yml

##### ADDITIONAL INFORMATION
Additional bug fixes to the test cases.  Removing invalid or deprecated options.

### DIFF
--- a/plugins/modules/azure_rm_rediscache.py
+++ b/plugins/modules/azure_rm_rediscache.py
@@ -71,6 +71,42 @@ options:
             - When set I(enable_non_ssl_port=true), the non-ssl Redis server port 6379 will be enabled.
         type: bool
         default: false
+    identity:
+        description:
+            - Identity for the WebApp.
+        type: dict
+        version_added: '2.7.0'
+        suboptions:
+            type:
+                description:
+                    - Type of the managed identity
+                choices:
+                    - SystemAssigned
+                    - UserAssigned
+                    - SystemAssigned, UserAssigned
+                    - None
+                default: None
+                type: str
+            user_assigned_identities:
+                description:
+                    - User Assigned Managed Identities and its options
+                required: false
+                type: dict
+                default: {}
+                suboptions:
+                    id:
+                        description:
+                            - List of the user assigned identities IDs associated to the WebApp
+                        required: false
+                        type: list
+                        elements: str
+                        default: []
+                    append:
+                        description:
+                            - If the list of identities has to be appended to current identities (true) or if it has to replace current identities (false)
+                        required: false
+                        type: bool
+                        default: True
     maxfragmentationmemory_reserved:
         description:
             - Configures the amount of memory in MB that is reserved to accommodate for memory fragmentation.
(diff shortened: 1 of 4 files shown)


tests/integration/targets/azure_rm_rediscache/tasks/main.yml
```diff
@@ -489,6 +529,22 @@
...
+    resource_group: "{{ resource_group }}"
+    provider: ManagedIdentity
+    resource_type: userAssignedIdentities
+    resource_name: "{{ item }}"
+    api_version: "2023-01-31"
+    state: absent
+  loop:
+    - "ansible-test-redis-identity"
+    - "ansible-test-redis-identity-2"
+
```
Fred-sun: Too many blank lines
```suggestion
```

plugins/modules/azure_rm_rediscache.py
```diff
@@ -71,6 +71,42 @@
...
+                        required: false
+                        type: list
+                        elements: str
+                        default: []
+                    append:
+                        description:
+                            - If the list of identities has to be appended to current identities (true) or if it has to replace current identities (false)
+                        required: false
+                        type: bool
+                        default: False
```
Fred-sun:
```suggestion
                        default: True
```

Fred-sun (commented):
small change request!

plugins/modules/azure_rm_rediscache.py
p3ck: So in #1643 you requested the default for append be False.  Here you are requesting we change the default to True.  

Can we agree on a default for all the modules?

Fred-sun: @p3ck As @xuzhang3  said, if default is false, we override the current identity by default. Thank you!

p3ck: If you agree with @xuzhang3 then the default here should be False.  If the default is False we will overwrite and only append if the user sets this to True

Fred-sun: @p3ck Ok, I agree!

Fred-sun: @p3ck  doc-default-does-not-match-spec: Argument 'append' in argument_spec found in identity -> user_assigned_identities defines default as (True) but documentation defines default as (False)

Fred-sun (commented):
small change request!

plugins/modules/azure_rm_rediscache.py
Fred-sun: @p3ck The argument default value is ['true',](url https://github.com/ansible-collections/azure/blob/997e4de875fa12e99c9a064b614797a65c58677e/plugins/module_utils/azure_rm_common_ext.py#L37) Please keep consistent. Thank you!

Fred-sun (commented):
Small change requst!

plugins/modules/azure_rm_rediscache.py
p3ck: done

tests/integration/targets/azure_rm_rediscache/tasks/main.yml
Fred-sun: Too many blank lines, Please delete the blank lines!
```suggestion
```

p3ck: Thanks, Done.

xuzhang3 (approved):

PR #1670, opened 2024-08-12
Fix SDK call to use Configuration class

Yaish25491:
Until now, begin_create_or_update function was called with values directly which caused failure due to unexpected format, str instead of dict.
Changes to use Configuration class for the parameter's value

##### SUMMARY

Wrapped value and source parameters with Configuration  class in begin_create_or_update function

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm_postgresqlconfiguration

##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_postgresqlconfiguration.py
+++ b/plugins/modules/azure_rm_postgresqlconfiguration.py
@@ -74,6 +74,7 @@ id:
 
 try:
     from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common import AzureRMModuleBase
+    from azure.mgmt.rdbms.postgresql.models import Configuration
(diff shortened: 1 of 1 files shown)


nirarg:
@Fred-sun @xuzhang3 can you please review this?
The integration test `azure_rm_postgresqlserver` was failed before this fix and now it passed (executed locally)
This test has skip flag, do you think we should remove the skip flag from it?
Started CI test:
https://dev.azure.com/azclitools/public/_build/results?buildId=180896&view=results
(Probably the test is going to be skipped here, right?)

nirarg (approved):

xuzhang3 (approved):

PR #1715, opened 2024-09-17
Add new "azure_arc" Inventoy Plugin file

nirarg:
##### SUMMARY
This new Inventory Plugin allow ssh to machines connected through Azure ARC

This plugin uses azure CLI to perform the following actions:
* Retrieve Azure Arc-connected machines (az connectedmachine list)
* Generate an SSH config record for the machine (az ssh config) 

This pluging has the following assumptions:
* Azure CLI is installed on the local machine and logged in with the right user
* All the Machines under the resource group configured with the same "username" and "ssh-key"

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
New Inventory Plugin:
azure.azcollection.azure_arc

### DIFF
--- a/plugins/inventory/azure_arc.py
+++ b/plugins/inventory/azure_arc.py
@@ -1,0 +1,193 @@
+# Copyright (c) 2018 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+
+__metaclass__ = type
+
+import json
+import os
+import subprocess
+
+from ansible.plugins.inventory import BaseInventoryPlugin
+
+DOCUMENTATION = """
+    name: azure_arc_ssh
+    short_description: Azure Arc inventory plugin using az ssh config for SSH connection
+    description:
+        - Fetches Azure Arc-connected machines and generates SSH config using az ssh config command.
+    options:
+        subscription_id:
+            description: Azure Subscription ID
+            required: false
+        resource_group:
+            description: Azure Resource Group where Arc machines are located
+            required: true
+        username:
+            description: Username for SSH connection
+            required: true
+        config_file_path:
+            description: Path where the SSH config file should be generated
+            required: false
+            default: /tmp/azure_arc_ssh_config
+        private_key_file:
+            description: Path to the SSH private key file
+            required: false
+            default: ~/.ssh/id_rsa
+"""
+# *** Consider adding auth:
+# client_id:
+#     description: Azure Client ID (App ID)
+#     required: false
+#     env:
+#       - name: AZURE_CLIENT_ID
+# client_secret:
+#     description: Azure Client Secret
+#     required: false
+#     env:
+#       - name: AZURE_CLIENT_SECRET
+# tenant_id:
+#     description: Azure Tenant ID
+#     required: false
+#     env:
+#       - name: AZURE_TENANT_ID
+
+
+class InventoryModule(BaseInventoryPlugin):
+    NAME = "azure.azcollection.azure_arc"
+
+    def verify_file(self, path):
+        """
+        Verify that file is usable by this plugin.
+        :param path: the path to the inventory config file
+        """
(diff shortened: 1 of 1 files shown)


nirarg:
I opened this PR in order to show and discuss the way I found to implement this
@Fred-sun @xuzhang3  @p3ck 
Please share your thoughts about this implementation here

Open questions:
* Is it ok to use the CLI here? can we have the assumption and requirement to have the azure CLI pre-installed on the local machine?
* Is there a way to install azure CLI on AAP execution environment?
* We currently can't install the CLI using "pip install", as result of dependency conflicts between the CLI and the ansible collection. Currently it can be installed only by the package manager. Is this ok?
* Is it ok to have the requirement that all machines are created with the same username and ssh key? any other thoughts/suggestions here?

xuzhang3:
@nirarg
1.  azcollection support CLI login so user can use CLI here.  What confused is why not using the Arc list APIs but the CLI command to get the machines?
2. Not sure if the AAP environment is isolated. CLI supports installation from package in local environment.
3. Recommend install the azcollection dependencies and CLI in virtual env to avoid the conflicts
4. Can we set it as default behavior to generate the same username and SSH key for all machines, but also support custom username and SSH key for each machine?

nirarg:
> azcollection support CLI login so user can use CLI here. What confused is why not using the Arc list APIs but the CLI command to get the machines?

We can likely use the API for the ARC Machine list, but I believe the CLI should still be used for the 'az ssh config' command. This is because most of the code for that functionality is located within the CLI itself and not in a shared library.

> Recommend install the azcollection dependencies and CLI in virtual env to avoid the conflicts

Yes, this solution can be used.
Is there a plan to upgrade the collection dependencies to resolve this conflict?

> Can we set it as default behavior to generate the same username and SSH key for all machines, but also support custom username and SSH key for each machine?

Currently, it searches for all machines within the same resource group.
To address this, the machines could either be separated into different resource groups, or we could add an additional search option, allowing them to be defined as different host records in the Inventory file.

xuzhang3:
> > azcollection support CLI login so user can use CLI here. What confused is why not using the Arc list APIs but the CLI command to get the machines?
> 
> We can likely use the API for the ARC Machine list, but I believe the CLI should still be used for the 'az ssh config' command. This is because most of the code for that functionality is located within the CLI itself and not in a shared library.

Not familiar with the CLI codes. We can take 'az ssh config' as a workaround but this also means we totally depends on the CLI implementation. It would be better if we supported this natively

> 
> > Recommend install the azcollection dependencies and CLI in virtual env to avoid the conflicts
> 
> Yes, this solution can be used. Is there a plan to upgrade the collection dependencies to resolve this conflict?

Working on it,  we are targeting on upgrade the dependencies recently. Some of the PR has been merged.

> 
> > Can we set it as default behavior to generate the same username and SSH key for all machines, but also support custom username and SSH key for each machine?
> 
> Currently, it searches for all machines within the same resource group. To address this, the machines could either be separated into different resource groups, or we could add an additional search option, allowing them to be defined as different host records in the Inventory file.

👍

plugins/inventory/azure_arc.py
```diff
@@ -0,0 +1,193 @@
...
+            "json",
+        ]
+
+        if subscription_id is not None:
+            command = command + ["--subscription", subscription_id]
+
+        try:
+            result = subprocess.run(command, capture_output=True, check=True, text=True)
+            machines = json.loads(result.stdout)
+            return [
```
ygalblum, line 142: Why are you returning a list of dictionaries that each only have one key? Why not just return the list of names?

ygalblum (commented):
Small comment

ygalblum:
> Not familiar with the CLI codes. We can take 'az ssh config' as a workaround but this also means we totally depends on the CLI implementation. It would be better if we supported this natively

From what @nirarg and I saw, it seems that most of the functionality is implemented in the client. So, implementing it natively will basically require copying a large chunk of code and maintaining it. So, yes, it would be better, but it doesn't seem feasible or worth the effort

p3ck:
This is superseded by #1735

PR #1833, opened 2025-02-26
[Enhance] `azure_rm_manageddisk` and `azure_rm_manageddisk_info` - Add support for more parameters manageddisk

### LINKED ISSUE #1829: azure_rm_manageddisk: allow creation of managed disk from a compute gallery image
##### SUMMARY
We are creating our own OS images and store them in an Azure Compute Gallery.  When then creating a VM we want to use that images. So, please add an option to allow creation of a managed disk from these compute gallery images.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_manageddisk

##### ADDITIONAL INFORMATION
More details about creation from the Azure API:
[Create a managed disk from an Azure Compute Gallery image](https://learn.microsoft.com/en-us/rest/api/compute/disks/create-or-update?view=rest-compute-2023-04-02&tabs=HTTP#create-a-managed-disk-from-an-azure-compute-gallery-image.)

```yaml
- azure.azcollection.azure_rm_manageddisk:
    name: disk1
    create_option: from_image
    gallery_image_reference:
       id: ...
```

Fred-sun:
##### SUMMARY

Add support for more parameters manageddisk, try to fixes #1829 

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_manageddisk.py
azure_rm_manageddisk_info.py
##### ADDITIONAL INFORMATION

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_manageddisk.py
+++ b/plugins/modules/azure_rm_manageddisk.py
@@ -76,10 +76,7 @@ options:
             - upload
             - fromimage
             - restore
-            - copystart
-            - importsecure
             - uploadpreparedsecure
-            - copyfromsansnapshot
     storage_account_id:
         description:
             - The full path to the storage account the image is to be imported from.
@@ -261,11 +258,6 @@ options:
         description:
             - If I(create_option=importsecure), this is the URI of a blob to be imported into VM guest state.
         type: str
-    elastic_san_resource_id:
-        description:
-            - Required if I(create_option=copyfromsansnapshot).
-            - This is the ARM id of the source elastic san volume snapshot.
-        type: str
     source_resource_id:
         description:
             - If I(create_option=copy), this is the ARM id of the source snapshot or disk.
@@ -343,7 +335,6 @@ EXAMPLES = '''
   azure_rm_manageddisk:
     resource_group: "{{ resource_group }}"
     name: "md{{ rpfx }}"
-    #storage_account_type: "UltraSSD_LRS"
     storage_account_type: "Standard_LRS"
     disk_size_gb: 1024
     network_access_policy: DenyAll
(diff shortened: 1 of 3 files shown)


cmasopust:
Had a quick look at the changes, especially for creating a managed disk from a gallery image.  Don't think that the current implementation is correct in this case.

From the example [Create a managed disk from an Azure Compute Gallery image](https://learn.microsoft.com/en-us/rest/api/compute/disks/create-or-update?view=rest-compute-2023-04-02&tabs=HTTP#create-a-managed-disk-from-an-azure-compute-gallery-image.) I see that it is sufficient to specify the `gallery_image_reference`, so the following should work (IMHO):

```yaml
    - name: create managed disk
      azure.azcollection.azure_rm_manageddisk:
        ...
        name: test-managed-disk-from-image_reference
        ...
        create_option: fromimage
        gallery_image_reference:
          id: /subscriptions/bla...
```

but it doesn't. According to the current implementation, a `security_profile` would be required as well, although the API doesn't require it:

```
TASK [create managed disk] **********************************************************************************************************************************************
fatal: [localhost]: FAILED! => {"changed": false, "msg": "create_option is fromimage but all of the following are missing: security_profile"}
```

cmasopust:
Another change that I'm not sure if it's correct is at line 1062:

```
disk_params['hyper_v_generation'] = 'V2'
```

As it's hardcoded to V2 in any case, does this mean that we no longer can create managed disks for hyper-v generation V1 ?

So the creation of a managed disk from a gallery image that is for V1 will fail?

Fred-sun:
@cmasopust 

> Another change that I'm not sure if it's correct is at line 1062:
> 
> ```
> disk_params['hyper_v_generation'] = 'V2'
> ```
> 
> As it's hardcoded to V2 in any case, does this mean that we no longer can create managed disks for hyper-v generation V1 ?
> 
> So the creation of a managed disk from a gallery image that is for V1 will fail?

@cmasopust I am still testing this PR,  Line 1062 is a test statement, I forgot to delete when uploading!  Already update, can you retry it?

Fred-sun:
@cmasopust Any feedback?

cmasopust:
Sorry, was in a meeting...

Yes, by removing the dependency for the security profile I was then able to create the managed disk from my gallery image!

Thanks!

Klaas-:
Regression in #1858

PR #1943, opened 2025-05-26
Add support tag module

### LINKED ISSUE #944: Azure Tagging
##### SUMMARY

Nice to have an ansible script that will be able to list all tags, remove, add, edit all tags from a subscription subsequently for any of the listed resources etc.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

aztagginginfo module

##### ADDITIONAL INFORMATION

This would be used for any and all tagging inside of a given subscription and would greatly automate as well as list what's needed. 

```yaml

ansible-playbook -i "subscription_id," -m aztagginginfo -a "resource_group=testrg" 

ansible-playbook -i "subscription_id," -m aztagginginfo ### Defaults to listing tags for subscription_id
```

Fred-sun:
##### SUMMARY

Add support tag module, try to fixes #944

##### ISSUE TYPE

- New Module Pull Request

##### COMPONENT NAME

azure_rm_tags.py
azure_rm_tags_info.py
##### ADDITIONAL INFORMATION

```paste below

```


tests/integration/targets/azure_rm_tags/tasks/main.yml
```diff
@@ -0,0 +1,154 @@
...
+    tags:
+      key1: value1
+  register: output
+
+- name: Assert idempotent
+  ansible.builtin.assert:
+    that:
+      - not output.changed
+
+- name: Pause for 3 mimutes
```
magodo: Why do we need to pause 3 min again here?

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,345 @@
...
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: azure_rm_tags
+version_added: "3.5.0"
+short_description: Manage tags
+description:
+    - Create, update ,delete the tags.
```
magodo:
```suggestion
    - Create, update, delete the tags.
```

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,345 @@
...
+- name: Create a tag with tag_name
+  azure_rm_tags:
+    tag_name: testkey
+
+- name: Create a tag with tag_value
+  azure_rm_tags:
+    tag_name: testkey
+    tag_value: testvalue
+
+- name: Create a new tags with scope
```
magodo, line 53:
```suggestion
- name: Create new tags with scope
```

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,345 @@
...
+- name: Delete the tag with tag_name
+  azure_rm_tags:
+    tag_name: testkey
+    state: absent
+
+- name: Delete the tag with tag_value
+  azure_rm_tags:
+    tag_name: testkey
+    tag_value: testvalue
+    state: absent
```
magodo: When deleting a tag, the `tag_value` is not necessary?

magodo (changes requested):

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,233 @@
...
+        if self.state == 'present':
+            response = self.get_at_scope()
+            if response and response['properties'].get('tags'):
+                if self.tags:
+                    update_tags = self.tags_update(response['properties']['tags'], self.tags)
+
+                    if (self.operation == 'Merge' or self.operation == 'Replace') and update_tags:
+                        changed = True
+                        response = self.begin_update_at_scope(self.tags, self.operation)
+                    elif not update_tags and self.operation == 'Delete':
```
magodo: Instead of testing on `not update_tags`, shoulnd't we just test if the remote has non-empty tags, which is already checked at line 146?

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,233 @@
...
+                        changed = True
+                        response = self.begin_update_at_scope(self.tags, self.operation)
+            else:
+                if self.tags:
```
magodo, line 152: In this case, if the operation is Delete, it should be a no-change.

plugins/modules/azure_rm_tags.py
```diff
@@ -0,0 +1,233 @@
...
+
+        self.results = dict(
+            changed=False,
+            tag_info=dict()
+        )
+
+        super(AzureRMTags, self).__init__(self.module_arg_spec,
+                                          supports_tags=True,
+                                          supports_check_mode=True)
+
```
magodo, line 136: state `absent` shall conflict with `Operation`?

magodo (changes requested):
Thanks for this PR!
I have left some additional mostly minor comments that once addressed this should be good to merge 👍

plugins/modules/azure_rm_tags.py
Fred-sun: No conflict, I(state=absent) to clear tags,  I(operation=Delete) to remove tags contain the paramertes

plugins/modules/azure_rm_tags.py
Fred-sun: It has been updated. Add, delete and replace will only be completed under three state.

Fred-sun:
@magodo ready_for_review!

magodo (approved):

PR #1972, opened 2025-06-19
Cleanup ado.sh and set default for pr-pipeline to latest versions (python+ansible)

magodo:
##### SUMMARY

- pr-pipeline:
  - Set default version to latest for python and ansible
  - Explicitly set `pr: none` to disable auto triggering by PR
  - Add a new step ("Determine the target ansible test group number") to echo the active gruop number to guide the test validator to only inspect that matrix job for inspecting the test result
  - Removing `python.version` in matrix setting as it is all the same
- ado.sh:
  - Remove python setup steps as it is setup by the UsePythonVersion ADO task already (note that this effectively removed the capability to test python2, which should be fine per experience). Another reason to remove those python setup is that a local run of the script will mass up the local python installations.
  - Early quit if the module is specified and the current group doesn't contain the module
  - Ensure cloud-config-azure.ini is removed before exit

### DIFF
--- a/pr-pipelines.yml
+++ b/pr-pipelines.yml
@@ -2,7 +2,7 @@ parameters:
   - name: PYTHON_VER
     displayName: 'Python Version'
     type: string
-    default: "3.10"
+    default: "3.13"
     values:
       - "2.7"
       - "3.6"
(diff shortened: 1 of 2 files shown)


Fred-sun:
@magodo If it is python3.12, setuptools is not available in the default environment and needs to be installed separately
```e.g.
Traceback (most recent call last):
  File "/home/cloudtest/ansible/setup.py", line 5, in <module>
    from setuptools import find_packages, setup
ModuleNotFoundError: No module named 'setuptools'
```
If the ansible version is v2.15, Ansisible -lint detection prompts an error and the version is not supported
```e.g.
WARNING: Unable to find project. Cannot determine changes. All tests will be executed.
ERROR    ansible-lint requires ansible-core>=2.16.11 and current version is 2.15.13.post0

##[error]Bash exited with code '3'.

Fred-sun (approved):

PR #1981, opened 2025-06-25
Support permanently deletes the specified vaults

Fred-sun:
##### SUMMARY

Support permanently deletes the specified vaults

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_keyvault.py

### DIFF
--- a/plugins/modules/azure_rm_keyvault.py
+++ b/plugins/modules/azure_rm_keyvault.py
@@ -277,7 +277,8 @@ options:
         description:
             - Whether permanently deletes the specified vault. aka Purges the deleted Azure key vault
             - Default value is C(True)
-            - When I(is_purge_deleted) is specified, the I(location) has to be configured. If not configured, the default locatioin of the resource group will be used.
+            - When I(is_purge_deleted) is specified, the I(location) has to be configured.
+              If not configured, the default locatioin of the resource group will be used.
         type: bool
         default: False
     state:
--- a/plugins/modules/azure_rm_keyvault.py
+++ b/plugins/modules/azure_rm_keyvault.py
@@ -811,6 +811,8 @@ class AzureRMVaults(AzureRMModuleBaseExt):
         try:
             response = self.mgmt_client.vaults.delete(resource_group_name=self.resource_group,
                                                       vault_name=self.vault_name)
+            if isinstance(response, LROPoller):
+                response = self.get_poller_result(response)
         except Exception as e:
             self.log('Error attempting to delete the Key Vault instance.')
             self.fail("Error deleting the Key Vault instance: {0}".format(str(e)))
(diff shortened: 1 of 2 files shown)


plugins/modules/azure_rm_keyvault.py
```diff
@@ -273,6 +273,13 @@
                         description:
                             - Property to specify whether NRP will ignore the check if parent subnet has serviceEndpoints configured.
                         type: bool
+    is_purge_deleted:
+        description:
+            - Whether permanently deletes the specified vault. aka Purges the deleted Azure key vault
+            - Default value is C(True)
+            - When I(is_purge_deleted) need I(location) to be configured, If not configured, the default locatioin of the resource group.
```
magodo:
```suggestion
            - When I(is_purge_deleted) is specified, the I(location) has to be configured. If not configured, the default locatioin of the resource group will be used.
```

BTW, isn't the `location` always available for this resource?

plugins/modules/azure_rm_keyvault.py
```diff
@@ -273,6 +273,13 @@
                         description:
                             - Property to specify whether NRP will ignore the check if parent subnet has serviceEndpoints configured.
                         type: bool
+    is_purge_deleted:
+        description:
+            - Whether permanently deletes the specified vault. aka Purges the deleted Azure key vault
+            - Default value is C(True)
```
magodo, line 279: Can we change this to `False`? Otherwise, it is a breaking change?

magodo (changes requested):
Does the test also need to update?

plugins/modules/azure_rm_keyvault.py
Fred-sun: Of course, I will update it!

plugins/modules/azure_rm_keyvault.py
Fred-sun: No , the location required for purge

Fred-sun: Look for the exist name within  the location

tests/integration/targets/azure_rm_keyvault/tasks/main.yml
```diff
@@ -434,6 +434,9 @@
 # azure_rm_keyvault hsm setup and test
 #
 
+- name: Pause for 600 seconds
```
magodo: Deos this work?

magodo (changes requested):

tests/integration/targets/azure_rm_keyvault/tasks/main.yml
Fred-sun: No，This test unstable! Thanks!

Fred-sun: It has been deleted

magodo (approved):

PR #1994, opened 2025-07-02
Add support for generating inventory with a graph query

p3ck:
##### SUMMARY
New Azure Inventory Plugin Using Resource Graph and KQL for Improved Performance in Environments with hundreds of subscriptions.

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
plugins/doc_fragments/azure_kql.py
plugins/inventory/azure_kql.py
requirements.txt
tests/integration/targets/inventory_azure/playbooks/create_inventory_config.yml
tests/integration/targets/inventory_azure/playbooks/empty_inventory_config.yml
tests/integration/targets/inventory_azure/playbooks/setup.yml
tests/integration/targets/inventory_azure/playbooks/test_inventory_kql.yml
tests/integration/targets/inventory_azure/runme.sh
tests/integration/targets/inventory_azure/templates/kql.yml

##### ADDITIONAL INFORMATION
The original idea came from a customer which had written a custom python script to generate a static inventory file.  We expanded on that and made it into a proper inventory plugin.

While there is some overlap with the current azure_rm inventory plugin I think this is different enough to warrent a separate plugin.  The query format gives a lot more control to the user to be able to generate inventory files.

Integration tests have been updated to handle this new module as well.

### DIFF
--- a/plugins/inventory/azure_kql.py
+++ b/plugins/inventory/azure_kql.py
@@ -18,6 +18,9 @@ DOCUMENTATION = r'''
         - See https://learn.microsoft.com/en-us/azure/virtual-machines/resource-graph-samples?tabs=azure-cli
           for how to craft your own query.  The one requirement is that you need to provide inventory_hostname.
         - Requires a YAML configuration file whose name ends with 'azure_kql.(yml|yaml)'
+        - Be aware that currently Azure Resource Graph may not be consistent with the actual state of your
+          resources.  It can take up to 30 minutes for updates to propagate.  This applies both for resources
+          to appear and to dissapear.
 '''
 
 EXAMPLES = '''
--- a/plugins/inventory/azure_kql.py
+++ b/plugins/inventory/azure_kql.py
@@ -1,0 +1,302 @@
+# Copyright (c) 2018 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = r'''
+    name: azure_kql
+    version_added: "3.7.0"
+    short_description: Azure Resource Manager inventory plugin using Graph QL
+    extends_documentation_fragment:
+      - azure.azcollection.azure
+      - azure.azcollection.azure_kql
+      - constructed
+      - inventory_cache
+    description:
+        - Query VM details from Azure Resource Manager using Graph QL
+        - See https://learn.microsoft.com/en-us/azure/virtual-machines/resource-graph-samples?tabs=azure-cli
+          for how to craft your own query.  The one requirement is that you need to provide inventory_hostname.
+        - Requires a YAML configuration file whose name ends with 'azure_kql.(yml|yaml)'
+'''
+
+EXAMPLES = '''
+plugin: azure.azcollection.azure_kql
+
+graph_query: |-
+  Resources
+      | where type =~ 'microsoft.compute/virtualmachines'
+      | project vmId = tolower(tostring(id)),
+                inventory_hostname = name,
+                tags,
+                location,
+                resourceGroup,
+                osType = tostring(properties.storageProfile.osDisk.osType),
+                powerState = tostring(properties.extended.instanceView.powerState.displayStatus),
+                hostName = properties.osProfile.computerName,
+                subscription_id = subscriptionId
+      | join kind=inner (ResourceContainers
+          | where type=='microsoft.resources/subscriptions'
+          | extend subscription_name = name,
+                   subscription_id = subscriptionId,
+                   state = properties.state
+          | where state == 'Enabled'
+          | project subscription_name,
+                    subscription_id)
+            on subscription_id
+      | project-away subscription_id1
+      | join (Resources
+          | where type =~ 'microsoft.network/networkinterfaces'
+          | mv-expand ipconfig=properties.ipConfigurations
+          | project vmId = tolower(tostring(properties.virtualMachine.id)),
+                    privateIp = ipconfig.properties.privateIPAddress,
+                    publicIpId = tostring(ipconfig.properties.publicIPAddress.id)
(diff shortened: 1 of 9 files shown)


p3ck:
I'm not sure why I'm getting this error in the sanity test though.. And only for 3.11

```
Running sanity test "import" on Python 3.8
Running sanity test "import" on Python 3.9
Running sanity test "import" on Python 3.10
Running sanity test "import" on Python 3.11
ERROR: Found 1 import issue(s) on python 3.11 which need to be resolved:
ERROR: plugins/inventory/azure_kql.py:104:0: traceback: ModuleNotFoundError: No module named 'pandas'
```

Fred-sun:
Our test environment did not rely on pandas, so this error did not occur. If an error occurs in python3.11, does it mean that pandas is not included in the python3.11 standard library and needs to be installed by oneself?

Fred-sun:
Please add 'azure-mgmt-resourcegraph' and 'pandas' to requirements.txt. Thanks

plugins/inventory/azure_kql.py
```diff
@@ -0,0 +1,296 @@
...
+    - prefix: ""
+      separator: ""
+      key: location
+    - prefix: ""
+      separator: ""
+      key: powerState
+'''
+
+import re
+import pandas as pd
```
Fred-sun:
```suggestion
try:
    import pandas as pd
except ImportError:
    # Avoid sanity detection errors
    pass
```

Fred-sun (commented):
I'm not sure what caused it.  Although the sanity detection indicates that 'pandas' does not exist, but there is no problem when azuer_kql.py is executed.  So I want to temporarily solve the current problem in this way.  Thank you!

plugins/inventory/azure_kql.py
p3ck: I see now that the azure_rm module had to do the same thing.

Fred-sun: We had encountered a similar problem before and did that at that time!

Klaas-:
Hi,
I was using Graph to generate an inventory and it kept coming up with incorrect output. I notified MSFT of that and they said by design Graph is not a reliable source of information. I would suggest to not use Graph for collection of an inventory.

Greetings
Klaas

Klaas-:
For the MSFT people, the support case number was 2406170050002212

I'll quote from it for the red hat people:

"""
For a definition of what ARG should do, I can only refer to the official documents that states “Azure Resource Graph is an Azure service designed to extend Azure Resource Management by providing efficient and performant resource exploration.”
I think the keyword in that sentence is “efficient”. If ARG also performed data validation, it would not be efficient anymore. I quoted an article earlier that mentioned the full scan that ARG is doing to make sure it has all its properties. This scan can take 30 hours to become available due to the amount of data it needs to query. If it also had to validate the data it received, it would take way so much time that the information would be out of date by the time it was made available.

I think the answer to your question “Does Microsoft want to change that?” is “No, we don’t want to change how this works right now”. Andreea discussed the case with both ARG and Compute resource groups. They found the reason for the missing property but due to the design, there is nothing that can be changed without re-architecting some major Azure components and this is not something we’re ready for at this point. Maybe in the future, but for now we’ll have to live with validating the properties on the client side.
"""

In the case I was exploring I noticed that important fields for an inventory (in this case current powerstate) would silently fail and just show "" instead of the actual powerstate. I have since reverted to using the API directly. This, at least from the information in that case, is the only interface on which microsoft provides "guarantees" for information accuracy.

magodo:
@Klaas- Thanks for bringing this limitation, I can also add one more limitation of ARG: for newly created resources, it will need some time to propagate to ARG until they are queriable.

@p3ck Do you mind to put down some note/warning in the document of this new inventory, about the ARG limitations?

Klaas-:
It also goes the other way, just on Friday I had another case with long-deleted resources still showing up in Graph. I stopped monitoring the situation after 8h post-delete because my day was finished. The answer from support is simply "this is a known issue" -- so also in that area, don't rely on Graph. If you want a proper inventory you _have_ to use API until MSFT rethinks what Graph is and what it should guarantee to it's users

p3ck:
> @Klaas- Thanks for bringing this limitation, I can also add one more limitation of ARG: for newly created resources, it will need some time to propagate to ARG until they are queriable.
> 
> @p3ck Do you mind to put down some note/warning in the document of this new inventory, about the ARG limitations?

I will put something in the notes.  I also have an inquiry to the customer who is currently generating their inventory via ARG and if they are aware of this issue.

Klaas-:
the official Resource Graph answer is "for new resource it can take up to 30 hours to appear in graph" and same for deleted resources "it can take up to 30 hours until a deleted resource disappears from graph"

p3ck:
I've put a disclaimer in the inventory module.

Fred-sun:
pipeline test: https://dev.azure.com/azclitools/internal/_build/results?buildId=259862&view=logs&j=b6daf38b-8237-596a-54bf-319fae9744cb&t=4d624bf0-640b-531f-6c03-5be4c6767d72

magodo (approved):

plugins/inventory/azure_kql.py
```diff
@@ -0,0 +1,305 @@
...
+      - azure.azcollection.azure_kql
+      - constructed
+      - inventory_cache
+    description:
+        - Query VM details from Azure Resource Manager using Graph QL
+        - See https://learn.microsoft.com/en-us/azure/virtual-machines/resource-graph-samples?tabs=azure-cli
+          for how to craft your own query.  The one requirement is that you need to provide inventory_hostname.
+        - Requires a YAML configuration file whose name ends with 'azure_kql.(yml|yaml)'
+        - Be aware that currently Azure Resource Graph may not be consistent with the actual state of your
+          resources.  It can take up to 30 minutes for updates to propagate.  This applies both for resources
```
Klaas-, line 22: I think you misread my comment :) 30 _hours_ not minutes.

Klaas-:
I've made a fix and some additions to warning in #2002

PR #1996, opened 2025-07-04
Convert the certificate data of type bytearray to base64

### LINKED ISSUE #1995: Ansible module azure_rm_keyvaultcertificate_info returns bytearray for cert_data
When  you go to pull the information of a certificate via Ansible using:

azure_rm_keyvaultcertificate_info

In the return attributes, the field 'cert_data' is being returned as a bytearray type, which (to my reading) is not able to revert back to any usable format without passing it through a python script.  

[https://github.com/ansible-collections/azure/blob/635b67c8e7411f956885b689c701b7cebf6e46b5/plugins/modules/azure_rm_keyvaultcertificate_info.py#L77-L88](https://github.com/ansible-collections/azure/blob/635b67c8e7411f956885b689c701b7cebf6e46b5/plugins/modules/azure_rm_keyvaultcertificate_info.py#L77-L88)

Would it be better to return that data as base64 encoded data instead?  At least there's a way to convert it in Ansible if you need to do anything with it.  

More of a question than an issue - but I'm trying to store a cert, key, and chain in a key vault so I can pull it later and deploy to a number of servers/applications/etc.

Fred-sun:
##### SUMMARY

Convert the certificate data of type bytearray to base64, try to fixes #1995

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm_keyvaultcertificate.py
azure_rm_keyvaultcertificate_info.py

### DIFF
--- a/plugins/modules/azure_rm_keyvaultcertificate.py
+++ b/plugins/modules/azure_rm_keyvaultcertificate.py
@@ -278,7 +278,7 @@ certificates:
                 - CER contents of the X509 certificate.
             type: str
             returned: always
-            sample: "bytearray(b'0.....................x16')"
+            sample: "MIID*****************0pRjXE"
         name:
             description:
                 - The name of the certificate.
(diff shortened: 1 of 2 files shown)


Fred-sun:
test pipeline: https://dev.azure.com/azclitools/internal/_build/results?buildId=259531&view=results

magodo:
This LGTM, except can we introduce a new property `cert_data_base64` to not break existing user's playbooks?

magodo (approved):

PR #2002, opened 2025-07-08
Fix warning on graph inventory

Klaas-:
##### SUMMARY

I have fixed the warning, it's 30 hours not 30 minutes and also added a sentence about the accuracy expectations of Graph.

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

plugins/inventory/azure_kql.py

### DIFF
--- a/plugins/inventory/azure_kql.py
+++ b/plugins/inventory/azure_kql.py
@@ -19,8 +19,11 @@ DOCUMENTATION = r'''
           for how to craft your own query.  The one requirement is that you need to provide inventory_hostname.
         - Requires a YAML configuration file whose name ends with 'azure_kql.(yml|yaml)'
         - Be aware that currently Azure Resource Graph may not be consistent with the actual state of your
-          resources.  It can take up to 30 minutes for updates to propagate.  This applies both for resources
+          resources.  It can take up to 30 hours for updates to propagate.  This applies both for resources
           to appear and to dissapear.
+          Microsoft also makes no guarantees that the information contained is accurate. It happens that
+          attributes (such as the PowerState) or entire VMs simply have an incorrect state, disappear and/or
+          reappear in different queries.
 '''
 
 EXAMPLES = '''


p3ck (approved):
Thanks for the update.

p3ck:
It is interesting to me that they don't mention any delay in their docs...

https://learn.microsoft.com/en-us/azure/governance/resource-graph/overview#how-resource-graph-is-kept-current

Klaas-:
@p3ck yeah that was also a complaint of mine in my support cases, but MSFT does not seem to care. You can find hints about the 30 hours limit in some other docs though: https://learn.microsoft.com/en-us/azure/network-watcher/network-insights-topology :)

plugins/inventory/azure_kql.py
```diff
@@ -19,8 +19,11 @@
           for how to craft your own query.  The one requirement is that you need to provide inventory_hostname.
         - Requires a YAML configuration file whose name ends with 'azure_kql.(yml|yaml)'
         - Be aware that currently Azure Resource Graph may not be consistent with the actual state of your
-          resources.  It can take up to 30 minutes for updates to propagate.  This applies both for resources
+          resources.  It can take up to 30 hours for updates to propagate.  This applies both for resources
           to appear and to dissapear.
+          Microsoft also makes no guarantees that the information contained is accurate. It happens that
+          attributes (such as the PowerState) or entire VMs simply have an incorrect state, disappear and/or
+          reappear in different queries.
```
magodo, line 26: Is there a reference to support this part?

Klaas-: if you have access to support requests you can check 2406170050002212 -- but no, from my opinion that's not properly documented in the public graph docs, same as the potential 30 hour lag, I've complained about that, but it's nothing MSFT wants to fix apparently

magodo: I hope at least you can put some reference about how ARG works and why it can cause these issues, to make the material more sensible.. Otherwise, I don't think anyone will use this inventory.

Klaas-: yeah, that's the point I am making, the graph service is not usable for an inventory if you have the expectation that your inventory shows accurate & live data :)

p3ck: Yet we have a customer who is using it for inventory.   And so far my automated tests have not shown any issues.

Klaas-: My guess would be they are not sensitive enough to change. For example the stuff with the powerstate="" in Graph I only noticed because some VMs where not updated in their maintenance windows, but only super rarely. I started to export my inventory to git after each sync in order to see changes, that's when I caught up to the general problems. Sometimes a VM would be in it, sometimes it would not be in it, sometimes a Powerstate was "" instead of the actual powerstate. According to MSFT this is all "by design". If that kind of inventory is okay for a customer they can feel free to use it. But it needs to be with a big warning sign because it's not what an actual inventory should be, which is live & accurate information -- or an error.

magodo: If ARG is behaving inproperly and MSFT claimed it was by design, then in what scenario is ARG supposed to be used?

Klaas-: @magodo that's actually the same question I posed to MSFT and they failed to come up with an answer :) I am guessing it shall be used to situations where neither accuracy/completeness nor time is an immediate factor. The only answer I came up in my head: statistics over your whole environment when you have thousands of machines and it's not important if there is one or more in the results or if one is powered on or simply "" :D . But at least in my ansible world that's not an option, there are important tasks running that rely on the accuracy of the information.

PR #2017, opened 2025-07-16
Azure servicebus - add async functionality

ptoscano:
Forwarding https://github.com/ansible/event-driven-ansible/pull/451 to the version of the `azure_service_bus.py` EDA source in this collection; quoting from that PR:
> This PR is to fix the issue https://github.com/ansible/event-driven-ansible/issues/401 , in which the async functionality was not working as expected. I've rewritten this module to leverage the Azure Service Bus async functionality.
> 
> In my testing, this resolves the QueueFull message that was being received when a large number of events needed to be processed.

### DIFF
--- a/extensions/eda/plugins/event_source/azure_service_bus.py
+++ b/extensions/eda/plugins/event_source/azure_service_bus.py
@@ -1,10 +1,9 @@
 import asyncio
-import concurrent.futures
 import contextlib
 import json
 from typing import Any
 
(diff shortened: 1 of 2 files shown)


p3ck (approved):
Thank you for this!  And thank you for re-submitting the PR while we transition this plugin from the EDA repo to here.

susanhooks:
Thanks all!

PR #2030, opened 2025-07-31
Delete import get cli profile

Fred-sun:
##### SUMMARY

Delete import get cli profile

##### COMPONENT NAME

azure_rm_common.py

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -312,6 +312,7 @@ except ImportError:
 
 try:
     from azure.cli.core.util import CLIError
+    from azure.cli.core._profile import Profile
     from azure.common.cloud import get_cli_active_cloud
 except ImportError:
     HAS_AZURE_CLI_CORE = False
@@ -1806,11 +1807,11 @@ class AzureRMAuth(object):
     def _get_azure_cli_credentials(self, subscription_id=None, resource=None):
         cloud_environment = get_cli_active_cloud()
 
+        subscription_id = subscription_id or self._get_env('subscription_id')
         az_cli = AzureCliCredential()
 
-        # SubscriptionClient to get the current subscription details
-        sub_list = SubscriptionClient(az_cli).subscriptions.list()
-        subscription_id = [item.subscription_id for item in sub_list][0] or subscription_id or self._get_env('subscription_id')
+        # Get the current subscription details
+        subscription_id = Profile().get_subscription()['id']
 
         cli_credentials = {
             'credentials': az_cli,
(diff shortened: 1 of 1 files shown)


plugins/module_utils/azure_rm_common.py
```diff
@@ -1805,21 +1804,16 @@ def _get_msi_credentials(self, subscription_id=None, client_id=None, _cloud_envi
...
-
-        cred, subscription_id, tenant = profile.get_login_credentials(
-            subscription_id=subscription_id)
         cloud_environment = get_cli_active_cloud()
 
         az_cli = AzureCliCredential()
+
+        # SubscriptionClient to get the current subscription details
+        sub_list = SubscriptionClient(az_cli).subscriptions.list()
+        subscription_id = [item.subscription_id for item in sub_list][0] or subscription_id or self._get_env('subscription_id')
```
magodo: Shall we change the priority here?

plugins/module_utils/azure_rm_common.py
```diff
@@ -1805,21 +1804,16 @@ def _get_msi_credentials(self, subscription_id=None, client_id=None, _cloud_envi
...
+
         cli_credentials = {
-            'credentials': az_cli if self.is_ad_resource else cred,
```
magodo, line 1822: What does the `is_ad_resource` mean here? Is it safe to always use `az_cli` as the credentials?

magodo (changes requested):
Thank you for the PR! I've left some comments inline. Additionally, shall we also update the dependency of `azure-cli-core` to its latest version, and see if we can safely remove the dependency on `msrestazure`?

plugins/module_utils/azure_rm_common.py
Fred-sun: Yes,  Now, All module support the credentials of SDK2. Thank you!

plugins/module_utils/azure_rm_common.py
Fred-sun: It was originally the first one because I just combined these three sentences into one sentence.

Fred-sun:
> Thank you for the PR! I've left some comments inline. Additionally, shall we also update the dependency of `azure-cli-core` to its latest version, and see if we can safely remove the dependency on `msrestazure`?

I will update azure-cli-core in another PR and complete the test. Thank you!

Fred-sun:
@magodo Please help re-check those changes(use azure-cli-core). Thank you!

Fred-sun:
@magodo ready_for_review

magodo (approved):

PR #2039, opened 2025-08-18
Replacement will send raw.githubusercontent.com request method

Fred-sun:
##### SUMMARY

Replacement will send raw.githubusercontent.com request method

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1603,6 +1603,7 @@ class AzureRMAuth(object):
         else:
             self._fail_impl = self._default_fail_impl
         self.is_ad_resource = is_ad_resource
+        self.auth_source = auth_source
 
         # authenticate
         self.credentials = self._get_credentials(
@@ -1810,6 +1811,8 @@ class AzureRMAuth(object):
                 cmd = ["az", "account", "show", "--query", "id", "-o", "tsv"]
                 subscription_id = subprocess.run(cmd, capture_output=True, text=True, check=True).stdout.strip()
             except Exception as ec:
+                if self.auth_source == 'auto':
+                    return None
                 self.fail("Obtain the az login's subscription occurred exception, exception information {0}".format(ec))
 
         try:
@@ -1818,8 +1821,12 @@ class AzureRMAuth(object):
             all_clouds = [x[1] for x in inspect.getmembers(azure_cloud) if isinstance(x[1], azure_cloud.Cloud)]
             matched_clouds = [x for x in all_clouds if x.name == cloud_name]
             if len(matched_clouds) != 1:
+                if self.auth_source == 'auto':
+                    return None
                 self.fail("Obtain the active cloud failed, there is no matching cloud.")
         except Exception as ec:
+            if self.auth_source == 'auto':
+                return None
             self.fail("Obtain the az login's active cloud occurred exception, exception information {0}".format(ec))
 
         cli_credentials = {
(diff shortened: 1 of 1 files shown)


plugins/module_utils/azure_rm_common.py
```diff
@@ -312,8 +312,7 @@ def default_api_version(self):
 
 try:
     from azure.cli.core.util import CLIError
-    from azure.cli.core._profile import Profile
-    from azure.common.cloud import get_cli_active_cloud
+    from azure.cli.core import cloud as azure_cloud
```
magodo, line 315: If the line above failed to import, this line won't be run?

plugins/module_utils/azure_rm_common.py
```diff
@@ -1808,13 +1807,25 @@ def _get_azure_cli_credentials(self, subscription_id=None):
...
-                subscription_id = Profile().get_subscription_id()
+                cmd = ["az", "account", "show", "--query", "id", "-o", "tsv"]
+                subscription_id = subprocess.run(cmd, capture_output=True, text=True, check=True).stdout.strip()
             except Exception as ec:
                 self.fail("Obtain the az login's subscription occurred exception, exception information {0}".format(ec))
+
+        try:
+            cmd = ["az", "cloud", "show", "--query", "name", "-o", "tsv"]
+            cloud_name = subprocess.run(cmd, capture_output=True, text=True, check=True).stdout.strip()
+            all_clouds = [x[1] for x in inspect.getmembers(azure_cloud) if isinstance(x[1], azure_cloud.Cloud)]
```
magodo, line 1818: What is the `x` here, wondering why `x[1]`?

plugins/module_utils/azure_rm_common.py
```diff
@@ -1808,13 +1807,25 @@ def _get_azure_cli_credentials(self, subscription_id=None):
         subscription_id = subscription_id or self._get_env('subscription_id')
         if not subscription_id:
             try:
-                subscription_id = Profile().get_subscription_id()
+                cmd = ["az", "account", "show", "--query", "id", "-o", "tsv"]
```
magodo: Can you check in which version of `az` the `-o tsv` is introduced to avoid this won't work for users who have older `az`?

plugins/module_utils/azure_rm_common.py
Fred-sun: Our requirements.txt requires the installation of azure-cli-core. It will not fail to import. Even if the import fails, errors will still occur when obtaining cloud information later!

plugins/module_utils/azure_rm_common.py
Fred-sun: 1818 line try to obtain the azure_cloud.cloud instance, returned a tuple like ('AZURE_CHINA_CLOUD', <azure.cli.core.cloud. Cloud object at 0x7ec5408973a0>), So use x[1].
```e.g.
[x for x in inspect.getmembers(azure_cloud) if isinstance(x[1], azure_cloud.Cloud)] return as
[('AZURE_CHINA_CLOUD', <azure.cli.core.cloud.Cloud object at 0x7de55a61b520>), ('AZURE_GERMAN_CLOUD', <azure.cli.core.cloud.Cloud object at 0x7de55a61b580>), ('AZURE_PUBLIC_CLOUD', <azure.cli.core.cloud.Cloud object at 0x7de55a61ae00>), ('AZURE_US_GOV_CLOUD', <azure.cli.core.cloud.Cloud object at 0x7de55a61ad70>)]

plugins/module_utils/azure_rm_common.py
Fred-sun: Ok. I will check the version that supports the minimum '-o tsv'. If it is lower. I will report an error and prompt the error message!

Fred-sun:
@magodo

magodo (approved):

PR #2041, opened 2025-08-20
Delete 'azure-cli-core' in requirements.txt

Fred-sun:
##### SUMMARY

Delete 'azure-cli-core' in requirements.txt

##### COMPONENT NAME

azure_rm_common.py
plugins/lookup/azure_service_principal_attribute.py

### DIFF
--- a/plugins/lookup/azure_service_principal_attribute.py
+++ b/plugins/lookup/azure_service_principal_attribute.py
@@ -31,7 +31,6 @@ options:
     description:
       - Azure cloud environment, default is C(AzureCloud), other optioins is C('AzureChinaCloud') and C('AzureUSGovernment').
       - Another option C(AzureGermanCloud) has deprecated on October 29th 2021.
-      - >- U(https://learn.microsoft.com/en-us/java/api/com.azure.identity.azureauthorityhosts?view=azure-java-stable&utm_source=chatgpt.com).
 """
 
 EXAMPLES = """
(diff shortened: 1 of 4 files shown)


plugins/lookup/azure_service_principal_attribute.py
```diff
@@ -72,15 +71,16 @@ def run(self, terms, variables, **kwargs):
...
+                                 AzureChinaCloud='https://login.chinacloudapi.cn',
+                                 AzureUSGovernment='https://login.microsoftonline.us')
+        azure_cloud = self.get_option('azure_cloud_environment', 'AzureCloud')
 
         try:
             azure_credential_track2 = ClientSecretCredential(client_id=credentials['azure_client_id'],
                                                              client_secret=credentials['azure_secret'],
                                                              tenant_id=credentials['azure_tenant'],
-                                                             authority=_cloud_environment.endpoints.active_directory)
+                                                             authority=cloud_environment.get(azure_cloud))
```
p3ck: Why don't you use the active_directory attribute from module_utls.azure_cloud?  Why duplicate it?

Fred-sun: Sure. I'll modify this module first. The final changes have not been completed yet. It will be changed to the latter later. Thank you!

Fred-sun: We cannot directly reference azure_cloud.py because when loading the file under lookup, ansible does not load module_utils into os.path. This led to a failed import. So here it is directly defined in the dictionary. Thank you!
```e.g.
  File "/home/cloudtest/ansible/lib/ansible/template/__init__.py", line 824, in _lookup
    ran = instance.run(loop_terms, variables=self._available_variables, **kwargs)
  File "/home/cloudtest/.ansible/ansible_collections/azure/azcollection/plugins/lookup/azure_service_principal_attribute.py", line 76, in run
    _cloud_environment = azure_cloud.HARD_CODED_CLOUD_DICT[cloud_name]
                         ^^^^^^^^^^^
NameError: name 'azure_cloud' is not defined
fatal: [testhost]: FAILED! => {
    "msg": "An unhandled exception occurred while running the lookup plugin 'azure.azcollection.azure_service_principal_attribute'. Error was a <class 'NameError'>, original message: name 'azure_cloud' is not defined. name 'azure_cloud' is not defined"
}

PR #67, opened 2020-03-16
Add SKU type for azure_rm_virtualnetworkgateway

Fred-sun:
##### SUMMARY

Add SKU type for azure_rm_virtualnetworkgateway.
This PR associate with [#66063](https://github.com/ansible/ansible/pull/66063).

##### ISSUE TYPE

- Bugfix Pull Request

##### COMPONENT NAME

azure_rm_virtualnetworkgateway
##### ADDITIONAL INFORMATION

```paste below

```

PR #2051, opened 2025-09-03
Add SKU type for azure_rm_virtualnetworkgateway

R4153:
##### SUMMARY
Add support for more gateway SKUs.

I couldn’t find an existing issue related to this.  
I did find an older PR with a similar fix: https://github.com/ansible-collections/azure/pull/67  

I’m not sure whether the default value should be updated, since the old SKUs are still working at the moment, even though Microsoft’s documentation suggests otherwise. 

Tested locally with a copy of the collection and confirmed it works.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
azure.azcollection.azure_rm_virtualnetworkgateway

##### ADDITIONAL INFORMATION
https://learn.microsoft.com/en-us/azure/vpn-gateway/gateway-sku-consolidation

```paste below

```

### DIFF
--- a/plugins/modules/azure_rm_virtualnetworkgateway.py
+++ b/plugins/modules/azure_rm_virtualnetworkgateway.py
@@ -104,7 +104,7 @@ options:
     sku:
         description:
             - The reference of the VirtualNetworkGatewaySku resource which represents the SKU selected for Virtual network gateway.
-        default: VpnGw1AZ
+        default: VpnGw1
         type: str
         choices:
             - VpnGw1
--- a/plugins/modules/azure_rm_virtualnetworkgateway.py
+++ b/plugins/modules/azure_rm_virtualnetworkgateway.py
@@ -278,7 +278,11 @@ class AzureRMVirtualNetworkGateway(AzureRMModuleBase):
             vpn_type=dict(type='str', default='route_based', choices=['route_based', 'policy_based']),
             vpn_gateway_generation=dict(type='str', default='Generation1', choices=['None', 'Generation1', 'Generation2']),
             enable_bgp=dict(type='bool', default=False),
-            sku=dict(type='str', default='VpnGw1AZ', choices=['VpnGw1', 'VpnGw2', 'VpnGw3', 'VpnGw1AZ', 'VpnGw2AZ', 'VpnGw3AZ', 'VpnGw4AZ', 'VpnGw5AZ', 'Standard', 'Basic', 'HighPerformance']),
+            sku=dict(type='str', default='VpnGw1', choices=[
+                'VpnGw1', 'VpnGw2', 'VpnGw3', 'Standard', 'Basic', 'HighPerformance',
+                'VpnGw1AZ', 'VpnGw2AZ', 'VpnGw3AZ', 'VpnGw4AZ', 'VpnGw5AZ',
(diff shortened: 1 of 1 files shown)


plugins/modules/azure_rm_virtualnetworkgateway.py
```diff
@@ -104,12 +104,17 @@
     sku:
         description:
             - The reference of the VirtualNetworkGatewaySku resource which represents the SKU selected for Virtual network gateway.
-        default: VpnGw1
+        default: VpnGw1AZ
```
Fred-sun: Do not change the default value unless necessary, as this will affect the current user.
```suggestion
        default: VpnGw1
```

plugins/modules/azure_rm_virtualnetworkgateway.py
```diff
@@ -273,7 +278,7 @@ def __init__(self):
             vpn_type=dict(type='str', default='route_based', choices=['route_based', 'policy_based']),
             vpn_gateway_generation=dict(type='str', default='Generation1', choices=['None', 'Generation1', 'Generation2']),
             enable_bgp=dict(type='bool', default=False),
-            sku=dict(type='str', default='VpnGw1', choices=['VpnGw1', 'VpnGw2', 'VpnGw3', 'Standard', 'Basic', 'HighPerformance']),
+            sku=dict(type='str', default='VpnGw1AZ', choices=['VpnGw1', 'VpnGw2', 'VpnGw3', 'VpnGw1AZ', 'VpnGw2AZ', 'VpnGw3AZ', 'VpnGw4AZ', 'VpnGw5AZ', 'Standard', 'Basic', 'HighPerformance']),
```
Fred-sun: The SDK also support  "VpnGw4", "VpnGw5", "UltraPerformance", "ErGw1AZ", "ErGw2AZ", "ErGw3AZ", and "ErGwScale" that haven't been added. Why not add them all together?

Fred-sun: line too long (193 > 160 characters)
```suggestion
            sku=dict(type='str', default='VpnGw1AZ', choices=['VpnGw1', 'VpnGw2', 'VpnGw3', 'VpnGw1AZ', 'VpnGw2AZ', 'VpnGw3AZ', 'VpnGw4AZ', 'VpnGw5AZ', 'Standard', 'Basic', 'HighPerformance']),
```

R4153: I didn’t test the ExpressRoute gateways, and I didn’t want to merge something I hadn’t tested. There could also be validation improvements regarding the generation, since VpnGw4AZ and VpnGw5AZ only work with Generation2 according to the table. However, that was too much for me to handle, as I’m not a Python developer. Furthermore, the older, non-AZ versions are no longer supported, so I don’t see any reason to include them. 
Source:
https://learn.microsoft.com/en-us/azure/vpn-gateway/gateway-sku-consolidation
https://learn.microsoft.com/en-us/azure/expressroute/gateway-migration

Fred-sun: Understood. I will submit a new PR test and add three other options. Thanks!

Fred-sun:
Pipeline test: https://dev.azure.com/azclitools/internal/_build/results?buildId=269921&view=results

magodo (approved):

PR #2081, opened 2025-10-13
Code default should reflect default provided by doc_fragment

TiTi:
##### SUMMARY
No functionnal change.
True value is coming from plugins\doc_fragments\azure_rm.py
Just put same value returned by `self.get_option('batch_fetch')` in the code (for comprehension).

##### ISSUE TYPE
- Docs Pull Request

##### COMPONENT NAME
inventory azure_rm.py

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -231,7 +231,7 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
 
         self.azure_auth = None
 
-        self._batch_fetch = False
+        self._batch_fetch = True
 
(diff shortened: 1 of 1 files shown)


zunyangc:
Thank you for your contribution. 

After reviewing the changes, we noticed that they do not introduce any functional differences, as the value is already being derived from `self.get_option('batch_fetch')`.
For now, we will not proceed with merging this request. However, if you have a specific use case or scenario that highlights the necessity of this change, please feel free to share it so we can discuss further.

zunyangc:
Since this Pull Request does not make any functional change or improvements, we will close this for now. Thanks!

PR #2096, opened 2025-11-03
Fix regression when using MSI

p3ck:
##### SUMMARY
get_option does not support a default value.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/lookup/azure_keyvault_secret.py

### DIFF
--- a/plugins/lookup/azure_keyvault_secret.py
+++ b/plugins/lookup/azure_keyvault_secret.py
@@ -176,14 +176,17 @@ class LookupModule(LookupBase):
         self.set_options(direct=kwargs)
 
         ret = []
-        vault_url = self.get_option('vault_url', None)
-        use_msi = self.get_option('use_msi', True)
+        vault_url = self.get_option('vault_url')
+        use_msi = self.get_option('use_msi')
+        # If use_msi is not defined default to True.
+        if use_msi is None:
+            use_msi = True
         TOKEN_ACQUIRED = False
         token = None
 
         token_params = {
             'api-version': '2018-02-01',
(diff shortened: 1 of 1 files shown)


zunyangc:
**Locally Validated MSI Authentication:**

_Attached screenshot shows the playbook and its successful result._
Playbook:
image

Result:
Screenshot 2025-11-04 154652

Validation Steps:
1. Created an Azure VM.
2. Added a User Managed Identity to the VM.
3. Created a Key Vault and a secret.
4. Assigned Key Vault Secrets Officer role to the User Managed Identity.
5. SSH into the VM and ran the playbook.

Observation:
Without explicitly setting `use_msi`, the plugin successfully fetched the secret via MSI as expected.

zunyangc:
Hi @magodo , validation pass, please proceed to merge.

zunyangc (approved):

magodo (approved):

vv0s4k:
@p3ck Thank you for fixing this. 
@zunyangc Just a note regarding validation based on my own experience. You might have missed one case. The plugin works well with single identity but if there are two identities assigned to a VM, for instance the system managed identity for the AADSSHLoginForLinux extension to work and a user managed identity for access to a key vault, the lookup fails. I assume it happens cause the plugin does not honor the AZURE_CLIENT_ID env variable, nor it provides any other ways of specifying what identity to use, hence it defaults to system managed one while only user assigned identity is allowed to pull secrets from KV. I hope this makes sense... 
Please do not take above as plain truth as this is kinda my assumption. I do not have expertise in the source code.

zunyangc:
Noted. Thanks @vv0s4k , will mark this down.

PR #2109, opened 2025-11-13
Fixes #2105: azure_rm_storageaccount added support to handle resource instance rules

### LINKED ISSUE #2105: azure_rm_storageaccount support for network resource instances
##### SUMMARY

Azure has added 'Resource Instances' to the Public Network Access tab.  This allows the entry of a tenant ID and a resource ID in the command section that previously only supported ip addresses for authorization.

##### ISSUE TYPE
- New options in command set are shown here.   https://learn.microsoft.com/en-us/cli/azure/storage/account/network-rule?view=azure-cli-latest#az-storage-account-network-rule-add
- Examples of the option in use https://learn.microsoft.com/en-us/azure/storage/common/storage-network-security-resource-instances?tabs=azure-cli

##### COMPONENT NAME

in CLI this has been placed in the equivalent section of ip_rules as used in azure_rm_storageaccount 

##### ADDITIONAL INFORMATION

This is coming into play with the backup of flexible postgress to backup vaults where the vault needs to be authorized to this storage account.  This allows access to the storage account with select networks enabled with a default behavior of deny.  It also bypasses the need to know of an IP bases subnet to authorize.

In the JSON the links appear under
"networkAcls": {
 "resourceAccessRules": [
{
"resourceId":
"tenantId":
] }

It is also worth noting differences in commandline vs console.  Console asks for TYPE and then shows you the resources of that type in your current tenant.  The command suggests this may someday support ID's in other tenants.

It looks like we would need to have a state for each entry of present or absent;

```yaml
resourceAccessRules:
  - resourceId:
    tenantId:
    state: present


```

zunyangc:
##### SUMMARY
- Fixes issue #2105  where azure_rm_storageaccount.network_acls did not support Azure Storage resource instance network rules (resourceAccessRules), preventing configuration of resource-based firewall exceptions.
- Enhance azure_rm_storageaccount to correctly handle resource instance rules for supported Azure services (e.g., Backup Vaults, Synapse workspaces, Event Hub namespaces).

- Key changes:
  - Added new suboption under network_acls
  - Implemented SDK model mapping for NetworkRuleSet.resource_access_rules
  - Updated idempotency logic
  - Updated module documentation
  - Integration tests added under tests/integration/targets/azure_rm_storageaccount
  - Updated azure_rm_storageaccount_info to return resource instance rules in facts.

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_storageaccount.py
- plugins/modules/azure_rm_storageaccount_info.py

### DIFF
--- a/plugins/modules/azure_rm_storageaccount.py
+++ b/plugins/modules/azure_rm_storageaccount.py
@@ -178,11 +178,14 @@ options:
                         description:
                             - The complete path to the subnet.
                         type: str
+                        required: true
                     action:
                         description:
                             - The only logical I(action=Allow) because this setting is only accessible when I(default_action=Deny).
                         default: 'Allow'
                         type: str
+                        choices:
+                            - Allow
             ip_rules:
                 description:
                     - A list of IP addresses or ranges in CIDR format.
@@ -193,11 +196,14 @@ options:
                         description:
                             - The IP address or range.
                         type: str
+                        required: true
                     action:
                         description:
                             - The only logical I(action=Allow) because this setting is only accessible when I(default_action=Deny).
                         default: 'Allow'
                         type: str
+                        choices:
+                            - Allow
             resource_access_rules:
                 description:
                     - List of resource instance rules that allow specific Azure resources to access storage account when public network access is restricted.
@@ -210,13 +216,13 @@ options:
                     resource_id:
                         description:
                             - The full ARM resource ID of the Azure resource instance.
-                            - Must start with C(/subscriptions/).
                         type: str
                         required: true
                     tenant_id:
                         description:
-                            - The Azure Active Directory tenant ID of the resource instance. Defaults to the current tenant id.
+                            - The Azure Active Directory tenant ID of the resource instance.
                         type: str
+                        required: true
     blob_cors:
         description:
             - Specifies CORS rules for the Blob service.
@@ -550,7 +556,6 @@ EXAMPLES = '''
       bypass: AzureServices
       resource_access_rules: []
 '''
-
 
 RETURN = '''
 state:
@@ -1008,14 +1013,28 @@ class AzureRMStorageAccount(AzureRMModuleBaseExt):
                 options=dict(
                     bypass=dict(type='str', default='AzureServices', no_log=False),
                     default_action=dict(type='str', choices=['Allow', 'Deny'], default='Allow'),
-                    virtual_network_rules=dict(type='list', elements='dict'),
-                    ip_rules=dict(type='list', elements='dict'),
+                    virtual_network_rules=dict(
+                        type='list',
+                        elements='dict',
+                        options=dict(
+                            id=dict(type='str', required=True),
+                            action=dict(type='str', choices=['Allow'], default='Allow'),
+                        ),
+                    ),
+                    ip_rules=dict(
+                        type='list',
+                        elements='dict',
+                        options=dict(
+                            value=dict(type='str', required=True),
+                            action=dict(type='str', choices=['Allow'], default='Allow'),
+                        ),
+                    ),
                     resource_access_rules=dict(
                         type='list',
                         elements='dict',
                         options=dict(
                             resource_id=dict(type='str', required=True),
-                            tenant_id=dict(type='str'),
+                            tenant_id=dict(type='str', required=True),
                         ),
                     ),
                 ),
@@ -1032,16 +1051,20 @@ class AzureRMStorageAccount(AzureRMModuleBaseExt):
                         type='dict',
                         options=dict(
                             blob=dict(
-                                type='dict', options=blob_spec
+                                type='dict',
+                                options=blob_spec
                             ),
                             table=dict(
-                                type='dict', options=table_spec
+                                type='dict',
+                                options=table_spec
                             ),
                             queue=dict(
-                                type='dict', options=queue_spec
+                                type='dict',
+                                options=queue_spec
                             ),
                             file=dict(
-                                type='dict', options=file_spec
+                                type='dict',
+                                options=file_spec
                             )
                         )
                     ),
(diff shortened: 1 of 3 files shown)


zunyangc:
Hi @magodo , this PR is ready for review and merge. Pipeline [test results](https://dev.azure.com/azclitools/internal/_build/results?buildId=283126&view=results) passed. Thank you.

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -198,6 +198,27 @@
                             - The only logical I(action=Allow) because this setting is only accessible when I(default_action=Deny).
                         default: 'Allow'
                         type: str
+            resource_access_rules:
+                description:
+                    - List of resource instance rules that allow specific Azure resources to access storage account when public network access is restricted.
+                    - Each entry must include the full ARM resource ID and the tenant ID.
```
magodo: This is already explained by the option structure, hencen can be deleted:

```suggestion
```

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -198,6 +198,27 @@
...
+                suboptions:
+                    resource_id:
+                        description:
+                            - The full ARM resource ID of the Azure resource instance.
+                            - Must start with C(/subscriptions/).
+                        type: str
+                        required: true
+                    tenant_id:
+                        description:
+                            - The Azure Active Directory tenant ID of the resource instance.
```
magodo, line 223:
```suggestion
                            - The Azure Active Directory tenant ID of the resource instance. Defaults to the current tenant id.
```

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -952,7 +1005,23 @@ def __init__(self):
             allow_shared_key_access=dict(type='bool'),
             allow_cross_tenant_replication=dict(type='bool'),
             default_to_o_auth_authentication=dict(type='bool'),
-            network_acls=dict(type='dict'),
+            network_acls=dict(
+                type='dict',
+                options=dict(
+                    bypass=dict(type='str', default='AzureServices', no_log=False),
+                    default_action=dict(type='str', choices=['Allow', 'Deny'], default='Allow'),
+                    virtual_network_rules=dict(type='list', elements='dict'),
```
magodo: Shall we specify the `options` here as well?

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -952,7 +1005,23 @@ def __init__(self):
...
+                    ip_rules=dict(type='list', elements='dict'),
```
magodo: Shall we specify the `options` here as well?

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -1176,7 +1241,6 @@ def account_obj_to_dict(self, account_obj, blob_mgmt_props=None, blob_client_pro
             default_to_o_auth_authentication=account_obj.default_to_o_auth_authentication,
             allow_cross_tenant_replication=account_obj.allow_cross_tenant_replication,
             allow_shared_key_access=account_obj.allow_shared_key_access,
-            network_acls=account_obj.network_rule_set,
```
magodo, line 1179: It seems the model of the API matches this plugin, why do we need a special handling at L1341?

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -1715,12 +1841,50 @@ def update_static_website(self):
...
+                    ]
+
+                resource_access_rules = None
+                if nacls_local.get('resource_access_rules'):
+                    resource_access_rules = [
+                        models.ResourceAccessRule(
+                            resource_id=r.get('resource_id'),
+                            tenant_id=r.get('tenant_id')
+                        ) for r in (nacls_local.get('resource_access_rules') or [])
+                    ]
```
magodo, line 1883: Same here, why do we need this explicit handling? If the previous code works fine, it means at least for the `ip_rules` and vnet rules, they don't need the explict handling above?

tests/integration/targets/azure_rm_storageaccount/tasks/main.yml
```diff
@@ -450,6 +451,40 @@
...
+        name: "{{ storage_account_name_default }}10"
+        account_type: Standard_LRS
+        kind: StorageV2
+        public_network_access: Enabled
+        network_acls:
+          default_action: Deny
+          bypass: AzureServices
+          resource_access_rules:
+            - resource_id: "/subscriptions/{{ azure_subscription_id }}/resourceGroups/{{ resource_group }}-storageaccount/providers/Microsoft.DataProtection/BackupVaults/testVault"
+              tenant_id: "{{ tenant_id }}"
```
magodo, line 466: Since the tenant id is optional, can we omit it here as it is the same tenant?

tests/integration/targets/azure_rm_storageaccount/tasks/main.yml
```diff
@@ -450,6 +451,40 @@
...
+            - resource_id: "/subscriptions/{{ azure_subscription_id }}/resourceGroups/{{ resource_group }}-storageaccount/providers/Microsoft.ApiManagement/service/testAPI"
+              tenant_id: "{{ tenant_id }}"
```
magodo, line 468: Same here

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -198,6 +198,26 @@
                             - The only logical I(action=Allow) because this setting is only accessible when I(default_action=Deny).
                         default: 'Allow'
                         type: str
+            resource_access_rules:
+                description:
+                    - List of resource instance rules that allow specific Azure resources to access storage account when public network access is restricted.
+                    - Supported resource types
+                      U(https://learn.microsoft.com/en-us/azure/storage/common/storage-network-security-trusted-azure-services?source=recommendations)
+                    - Resource instance rules only apply when I(default_action=Deny) and when I(public_network_access=Enabled).
+                    - These rules allow access to the public endpoint only; RBAC permissions are still required for data operations.
```
magodo:
```suggestion
```

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -198,6 +198,26 @@
...
+                type: list
+                elements: dict
+                suboptions:
+                    resource_id:
+                        description:
+                            - The full ARM resource ID of the Azure resource instance.
+                            - Must start with C(/subscriptions/).
```
magodo:
```suggestion
```

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -1274,10 +1337,17 @@ def account_obj_to_dict(self, account_obj, blob_mgmt_props=None, blob_client_pro
             if account_obj.identity:
                 account_dict['identity'] = account_obj.identity.as_dict()
```
magodo: These might need to be left shift one level...

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -1303,15 +1373,54 @@ def failover_account(self):
...
+                    resource_access_rules=resource_access_rules
+                )
+
+                parameters = self.storage_models.StorageAccountUpdateParameters(network_rule_set=nrs)
+                self.storage_client.storage_accounts.update(self.resource_group, self.name, parameters)
             except Exception as exc:
-                self.fail("Failed to update account type: {0}".format(str(exc)))
+                self.fail("Failed to update network ACLs: {0}".format(str(exc)))
 
-    def sort_list_of_dicts(self, rule_set, dict_key):
```
magodo, line 1313: This is not necessary as we can just rely on the "resource_id" for sorting, which is unique across tenants.

plugins/modules/azure_rm_storageaccount.py
```diff
@@ -198,6 +198,26 @@
...
+                    resource_id:
+                        description:
+                            - The full ARM resource ID of the Azure resource instance.
+                            - Must start with C(/subscriptions/).
+                        type: str
+                        required: true
+                    tenant_id:
+                        description:
+                            - The Azure Active Directory tenant ID of the resource instance. Defaults to the current tenant id.
+                        type: str
```
magodo, line 224:
```
env:
          - name: AZURE_TENANT_ID
```

magodo (changes requested):

zunyangc:
Hi @magodo , thank you very much for the feedbacks! Changes have been made. 
Pipeline test [passed](https://dev.azure.com/azclitools/internal/_build/results?buildId=283627&view=results).

magodo (approved):

PR #2119, opened 2025-11-28
Idempotency fix: Change IP configuration to construct list instead of set

zunyangc:
##### SUMMARY
Change IP configuration to construct list instead of set. ADO idempotency failure comes from using set(...) in `construct_ip_configuration_set`. `default_compare` is list-aware (it sorts lists) but it is not set-aware and ends up comparing stringified sets, which is order‑unstable.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_networkinterface.py

### DIFF
--- a/plugins/modules/azure_rm_networkinterface.py
+++ b/plugins/modules/azure_rm_networkinterface.py
@@ -925,13 +925,13 @@ class AzureRMNetworkInterface(AzureRMModuleBaseExt):
             private_ip_allocation_method=to_native(item.get('private_ip_allocation_method')),
             public_ip_address_name=(to_native(item.get('public_ip_address').get('name'))
                                     if item.get('public_ip_address') else to_native(item.get('public_ip_address_name'))),
-            load_balancer_backend_address_pools=(set([to_native(self.backend_addr_pool_id(id))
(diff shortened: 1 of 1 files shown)


zunyangc:
Hi @magodo , pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=284976&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_networkinterface.py
magodo, line 761: Do you also need to check this part? It seems the set is used for some reason?

zunyangc: It’s not a concern because this block doesn’t rely on set() —it simply compares two normalized lists returned by `construct_ip_configuration_set`. This loop only checks for missing IP configuration names and appends them if needed, so no additional changes are required. 

Reference: https://github.com/ansible-collections/azure/blob/7b7ba8976f04adcd39220bcca4d66fca9987a618/plugins/module_utils/azure_rm_common_ext.py#L208

magodo (approved):

PR #2133, opened 2025-12-09
Fix #1580 [Follow Up]: Support upload for bigger files

### LINKED ISSUE #1580: Add option for web file source to azure_rm_storageblob module
##### SUMMARY

Good day,
currently `azure_rm_storageblob` supports as a source option local file path only. I would like to request new feature for source - web URL. So file can be uploaded from web directly to storage account.

Similar feature is already present in AZ CLI:
`az storage blob copy start` has `source-uri` option to upload file from web to storage account

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME

azure_rm_storageblob

##### ADDITIONAL INFORMATION

It will be useful mainly for bigger files located on Internet, so they can be directly uploaded to storage account. I have a playbook which downloads OS images, which are then used to create VMs. Without this feature I have to either download them first to local drive and then upload (this needs lot of space) or use AZ CLI.

```yaml

```

zunyangc:
##### SUMMARY
Fix #1580. Support upload for bigger files.  

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_storageblob.py

##### ADDITIONAL INFORMATION
https://learn.microsoft.com/en-us/cli/azure/storage/blob/copy?view=azure-cli-latest#az-storage-blob-copy-start


zunyangc:
Hi @magodo, this is the follow up request from user in #1580. 
Reimplemented the `upload_blob_from_url` method to use [asynchronous copy](https://learn.microsoft.com/en-us/cli/azure/storage/blob/copy?view=azure-cli-latest#az-storage-blob-copy-start). 

Pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=287484&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_storageblob.py
```diff
@@ -604,28 +603,29 @@ def upload_blob_from_url(self):
...
-                client.upload_blob_from_url(
+        try:
+            client = self.blob_service_client.get_blob_client(container=self.container, blob=self.blob)
+            if not self.check_mode:
+                # Initiate asynchronous copy operation
+                copy_result = client.start_copy_from_url(
                     source_url=self.source_url,
-                    overwrite=self.force,
                     metadata=self.tags,
-                    content_settings=content_settings,
```
magodo, line 617: Does this mean the force and content_settings will not impact the upload behavior after this change?

zunyangc: Yes, switching to `start_copy_from_url` means `force` and `content_settings` no longer make an impact. The copy operation always overwrites an existing destination blob and does not accept content settings (similar to `az storage blob copy start` which [overwrites by default](https://learn.microsoft.com/en-us/python/api/azure-storage-blob/azure.storage.blob.blobclient?view=azure-python#azure-storage-blob-blobclient-start-copy-from-url) and [does not set content type during copy](https://learn.microsoft.com/en-us/cli/azure/storage/blob/copy?view=azure-cli-latest#az-storage-blob-copy-start)).

magodo: It sounds like a *dangerous* breaking change?

zunyangc: Got it. Here's what I changed in the [latest commit](https://github.com/ansible-collections/azure/pull/2133/commits/c1cb8bf6f39655a57b50a2fd76569ee79ec54b75) to keep it "non-breaking":
1. Overwrite control: I added a server-side precondition equivalent to AZ CLI's `--destination-if-none-match` by sending `If-None-Match: "*"`.
````
force=false (default) -> copy starts only if the destination doesn't exist (no overwrite, otherwise, skip silently and return changed=false. )
force=true -> copy proceeds and overwrites
````
2. Content settings: I store the requested `content_settings`, but they are not applied during copy (consistent with Azure CLI). Users can reapply them after the async copy completes via the existing `update_blob_content_settings` path (there are new feature request #2137 for retrieving the copy status, user can reapply the `content_settings`, after copy status=success).

File: plugins/modules/azure_rm_storageblob.py [hunk 2]
```diff
@@ -604,28 +607,53 @@ def upload_blob_from_url(self):
...
+                return
+            if not self.check_mode:
+                kwargs = {
+                    "source_url": self.source_url,
+                    "metadata": self.tags,
+                    "standard_blob_tier": self.get_blob_tier(self.standard_blob_tier),
+                }
+                # Server-side overwrite protection, equivalent to `--destination-if-none-match`
+                if not self.force:
+                    kwargs["if_none_match"] = "*"
```
magodo: This seems not necessary as reaching here indicates either the block doesn't exist or is in force mode.

plugins/modules/azure_rm_storageblob.py
```diff
@@ -604,28 +607,53 @@ def upload_blob_from_url(self):
...
+
+                # Start async copy
+                copy_result = client.start_copy_from_url(**kwargs)
+                copy_status = copy_result.get('copy_status', 'pending')
+                self.results['copy_status'] = copy_status
+                self.results['copy_id'] = copy_result.get('copy_id')
```
magodo: We need to define them in the `RETURN`.

plugins/modules/azure_rm_storageblob.py
```diff
@@ -604,28 +607,53 @@ def upload_blob_from_url(self):
...
+
+                # Defer HTTP headers if copy still pending
+                if content_settings and copy_status == 'pending':
+                    self.results['deferred_http_headers'] = True
```
magodo: What is this used for?

plugins/modules/azure_rm_storageblob.py
```diff
@@ -604,28 +607,53 @@ def upload_blob_from_url(self):
...
-                    metadata=self.tags,
-                    content_settings=content_settings,
-                    standard_blob_tier=self.get_blob_tier(self.standard_blob_tier)
+            self.blob_obj = self.get_blob()
+            self.results['changed'] = True
+            self.results['actions'].append(f'started copy of blob {self.blob} from {self.source_url}')
+            self.results['container'] = self.container_obj
+            self.results['blob'] = self.blob_obj
+            if content_settings:
+                self.results['deferred_http_headers'] = True
```
magodo: Also here, need clarification.

magodo (changes requested):
Can we introduce a new property, e.g. `wait` (defaults to true), to allow users to opt in/out whether to wait for the uploading finish before returning from this module. This requires a polling logic to poll the properties of the uploading task.

Reference:

- https://github.com/hashicorp/terraform-provider-azurerm/blob/3c7cbe7ffcbdffffcda412a08dfa7fbc5b09cabc/internal/services/storage/blobs.go#L58-L71
- https://github.com/hashicorp/terraform-provider-azurerm/blob/3c7cbe7ffcbdffffcda412a08dfa7fbc5b09cabc/internal/services/storage/blobs.go#L93-L106
- https://github.com/hashicorp/terraform-provider-azurerm/blob/3c7cbe7ffcbdffffcda412a08dfa7fbc5b09cabc/vendor/github.com/jackofallops/giovanni/storage/2023-11-03/blob/blobs/copy_and_wait.go#L12-L28

magodo (approved):

PR #2136, opened 2025-12-18
Fix #1208: Adding delete_option for data disk and os disk

### LINKED ISSUE #1208: azure_rm_virtualmachine option for auto delete all resources connected to created VM
I could not find a similar issue for azure_rm_virtualmachine

##### SUMMARY
creating a virtual machine using azure_rm_virtualmachine does not allow you to set flags on associated resources (like data/os disks, nics) that tell those resources to be deleted when the virtual machine itself is deleted. I would like functionality for this, similar to what has been added to. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_virtualmachine

##### ADDITIONAL INFORMATION
Cleaning up resources from deleted virtual machines, especially spot machines, is somewhat of a pain in the ass if the associated virtual machine has been created using azure_rm_virtualmachine (and thus without auto delete). At worst it is required to go through the azure portal to check which resources associated with the deleted VM were not removed alongside it. Adding this functionality will lower the amount of labor required to delete resources that should have the same lifetime as the VM, and will make it a lot more manageable for spot eviction handling

```yaml
- name: Create a VM with OS and multiple data managed disks -- autodelete
  azure_rm_virtualmachine:
    resource_group: myResourceGroup
    name: testvm001
    vm_size: Standard_D4
    managed_disk_type: Standard_LRS
    admin_username: "{{ username }}"
    ssh_public_keys:
      - path: /home/adminUser/.ssh/authorized_keys
        key_data: < insert your ssh public key here... >
    image:
      offer: 0001-com-ubuntu-server-focal
      publisher: canonical
      sku: 20_04-lts-gen2
      version: latest
    data_disks:
      - lun: 0
        managed_disk_id: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxxxxxx/resourceGroups/myResourceGroup/providers/Microsoft.Compute/disks/myDisk"
      - lun: 1
        disk_size_gb: 128
        managed_disk_type: Premium_LRS
        deleteOption: Delete <------- suggested addition

```

https://learn.microsoft.com/en-us/azure/virtual-machines/delete?tabs=rest2%2Ccli3%2Cportal4%2Cportal5 

I am unfamiliar with the API used, but the link above shows the functionality as part of the REST API/CLI/etc

zunyangc:
##### SUMMARY
Fix #1208. Add support for Azure `delete_option` on VM-attached disks, aligned with ARM behavior.
- OS disk: `os_disk_delete_option` (Delete | Detach) (Note: _Default to Detach if no specify, except ephemeral OS disks which must always be Delete_)
- Data disks: per-disk `delete_option` (Delete | Detach) (Note: _Default to Detach if no specify_)
- Default NIC behavior: Default NIC created by the VM (when no `network_interface_names` provided) are deleted with the VM)

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_virtualmachine.py

##### ADDITIONAL INFORMATION
- `remove_on_absent` remains available when VM is removed via ARM (state: absent). 
- Spot evictions trigger VM deletion outside Ansible; Azure applies the VM model's `delete_option` settings for OS and data disks.
- Network interfaces are handled according to Azure native behavior: default NICs created by the VM are removed with the VM; standalone or user-supplied NICs remain managed by the Network API (Future work: adding `delete_option` for `azure_rm_networkinterface.py`. 

References: https://learn.microsoft.com/en-us/rest/api/compute/virtual-machines/create-or-update?view=rest-compute-2025-02-01&tabs=HTTP#definitions


zunyangc:
Hi @magodo, pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=287711&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -349,6 +349,15 @@
...
                     - ReadWrite
+            delete_option:
+                description:
+                    - Specifies the delete behavior for the data disk when deletin the VM.
+                    - When set to C(Delete), the data disk is removed during VM deletion.
+                    - When set to C(Detach), the data disk is preserved.
+                type: str
+                choices:
+                    - Delete
+                    - Detach
```
magodo, line 362: Should we also specify `default: Detach` here?

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -644,6 +653,24 @@
...
+            - Detach
+    nic_delete_option:
+        description:
+            - Specifies the delete behavior for the VM's NIC when the VM is deleted.
+            - When set to C(Delete), the NIC will be removed automatically during VM deletion.
+            - When set to C(Detach), the NIC will be preserved.
+        type: str
+        choices:
+            - Delete
+            - Detach
```
magodo: Shall we also specify the `default` here?

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -1539,7 +1593,40 @@ def exec_module(self, **kwargs):
...
+                if self.os_disk_delete_option and self.os_disk_delete_option != current_osdisk_delete:
+                    self.log('CHANGED: OS disk delete_option differs.')
+                    differences.append('OS Disk delete_option')
+                    changed = True
+
+                # Data disks delete option diff
+                if self.data_disks and vm_dict['storage_profile'].get('data_disks'):
+                    for dd_arg in self.data_disks:
+                        if dd_arg.get('delete_option') is None:
+                            continue
```
magodo: This needs to be checked in case the remote has set the `delete_option` to Delete.

plugins/modules/azure_rm_virtualmachine.py
```diff
@@ -644,6 +653,24 @@
...
+    os_disk_delete_option:
+        description:
+            - Specifies the delete behavior for the VM's OS disk when the VM is deleted.
+            - When set to C(Delete), the OS disk will be removed automatically during VM deletion.
+            - When set to C(Detach), the OS disk will be preserved.
+        type: str
+        choices:
+            - Delete
+            - Detach
+    nic_delete_option:
```
magodo: Why isn't this a per NIC option?

magodo (changes requested):

plugins/modules/azure_rm_virtualmachine.py
zunyangc: Azure already treats deleteOption as Detach by default when the field is omitted in the VM model. If we set `default: Detach` in the module’s argument spec, the parameter becomes implicitly “set” even when the user doesn’t specify it. As a result, if a user manually changes `deleteOption` to Delete (e.g., in the portal/CLI) and later re-runs the playbook without explicitly setting `os_disk_delete_option` / `data_disks[].delete_option`, the module would revert it back to Detach.

And so I think a safer way is to document Detach as the Azure default, like [this](https://github.com/ansible-collections/azure/pull/2136/changes#diff-050afcc769a0920572289d3cf492f6aa974d9acbb0db79e61ad330ce9fab945fR356)
````
- Azure default to Detach if no specified.
````

**Comment by zunyangc** [comment]
Hi @magodo, pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=290377&view=results) passed. Ready to merge, thanks!

File: plugins/modules/azure_rm_virtualmachine.py [hunk 5]
```diff
@@ -1597,6 +1650,16 @@ def exec_module(self, **kwargs):
                     changed = True
                     vm_dict['storage_profile']['os_disk']['disk_size_gb'] = self.os_disk_size_gb
 
+                if self.os_disk_delete_option:
+                    cur_os_del = vm_dict['storage_profile']['os_disk'].get('delete_option')
+                    if cur_os_del is None:
+                        cur_os_del = 'Detach'  # Azure default
```
magodo: This is not a reliable assumption as the service behavior might change between api versions. Instead, if this is not specified and the user specified a value, we mark it as changed?

magodo (changes requested):

magodo (approved):

PR #2141, opened 2025-12-24
Added support for Elastic Cluster in azure_rm_postgresqlflexibleserver

zunyangc:
##### SUMMARY
Added elastic cluster support, enabling users to create and manage Citus-backed elastic clusters on Azure Database for PostgreSQL Flexible Server. 

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_postgresqlflexibleserver
- requirements.txt
- tests/integration/targets/azure_rm_postgresqlflexibleserver/tasks/main.yml

##### ADDITIONAL INFORMATION
References:
- https://learn.microsoft.com/en-us/rest/api/postgresql/servers/update
- https://learn.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/flexibleservers
- https://learn.microsoft.com/en-us/azure/postgresql/flexible-server/concepts-elastic-clusters

### DIFF
--- a/plugins/modules/azure_rm_postgresqlflexibleserver.py
+++ b/plugins/modules/azure_rm_postgresqlflexibleserver.py
@@ -845,7 +845,8 @@ class AzureRMPostgreSqlFlexibleServers(AzureRMModuleBaseExt):
                     update_fields = [
                         'sku', 'storage', 'cluster', 'backup', 'high_availability',
                         'maintenance_window', 'auth_config', 'identity', 'tags',
-                        'version', 'network', 'availability_zone', 'create_mode'
+                        'version', 'network', 'availability_zone', 'create_mode',
+                        'fully_qualified_domain_name'
                     ]
                     desired = {k: self.parameters.get(k) for k in update_fields}
                     current = {k: old_response.get(k) for k in update_fields}
--- a/plugins/modules/azure_rm_postgresqlflexibleserver.py
+++ b/plugins/modules/azure_rm_postgresqlflexibleserver.py
@@ -845,8 +845,7 @@ class AzureRMPostgreSqlFlexibleServers(AzureRMModuleBaseExt):
                     update_fields = [
                         'sku', 'storage', 'cluster', 'backup', 'high_availability',
                         'maintenance_window', 'auth_config', 'identity', 'tags',
-                        'version', 'network', 'availability_zone', 'create_mode',
-                        'fully_qualified_domain_name'
+                        'version', 'network', 'availability_zone', 'create_mode'
                     ]
                     desired = {k: self.parameters.get(k) for k in update_fields}
                     current = {k: old_response.get(k) for k in update_fields}
--- a/plugins/modules/azure_rm_postgresqlflexibleserver.py
+++ b/plugins/modules/azure_rm_postgresqlflexibleserver.py
@@ -81,7 +81,6 @@ options:
             - '15'
             - '16'
             - '17'
-            - '18'
     fully_qualified_domain_name:
         description:
             - The fully qualified domain name of a server.
@@ -712,7 +711,7 @@ class AzureRMPostgreSqlFlexibleServers(AzureRMModuleBaseExt):
             ),
             version=dict(
                 type='str',
-                choices=['11', '12', '13', '14', '15', '16', '17', '18']
+                choices=['11', '12', '13', '14', '15', '16', '17']
             ),
             fully_qualified_domain_name=dict(
                 type='str',
@@ -1053,12 +1052,9 @@ class AzureRMPostgreSqlFlexibleServers(AzureRMModuleBaseExt):
             result['auth_config']['tenant_id'] = item.auth_config.tenant_id
         else:
             result['auth_config'] = None
-        cluster_obj = getattr(item, 'cluster', None)
-        if cluster_obj is None and hasattr(item, 'properties'):
-            cluster_obj = getattr(item.properties, 'cluster', None)
-        if cluster_obj is not None:
-            result['cluster']['cluster_size'] = getattr(cluster_obj, 'cluster_size', None)
-            result['cluster']['default_database_name'] = getattr(cluster_obj, 'default_database_name', None)
+        if item.cluster is not None:
+            result['cluster']['cluster_size'] = item.cluster.cluster_size
+            result['cluster']['default_database_name'] = item.cluster.default_database_name
         else:
             result['cluster']['cluster_size'] = None
             result['cluster']['default_database_name'] = None
--- a/plugins/modules/azure_rm_postgresqlflexibleserver.py
+++ b/plugins/modules/azure_rm_postgresqlflexibleserver.py
@@ -1128,22 +1128,17 @@ class AzureRMPostgreSqlFlexibleServers(AzureRMModuleBaseExt):
             result['auth_config']['tenant_id'] = item.auth_config.tenant_id
         else:
             result['auth_config'] = None
-        cluster_obj = None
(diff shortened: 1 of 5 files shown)


zunyangc:
Hi @magodo, pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=288446&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -886,14 +964,14 @@ def update_postgresqlflexibleserver(self, body):
         self.log("Updating the PostgreSQL Flexible Server instance {0}".format(self.name))
         try:
             # structure of parameters for update must be changed
-            response = self.postgresql_flexible_client.servers.begin_update(resource_group_name=self.resource_group,
-                                                                            server_name=self.name,
-                                                                            parameters=body)
+            response = self.postgresql_flexible_client.servers.begin_create_or_update(resource_group_name=self.resource_group,
```
magodo, line 914: Is this change on purpose as it indicates to changing PATCH to PUT?

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -750,8 +820,8 @@ def exec_module(self, **kwargs):
                 setattr(self, key, kwargs[key])
             elif kwargs[key] is not None:
                 self.parameters[key] = kwargs[key]
-                for key in ['location', 'sku', 'administrator_login_password', 'storage', 'backup',
-                            'high_availability', 'maintenance_window', 'create_mode', 'auth_config']:
+                for key in ['location', 'sku', 'administrator_login_password', 'storage', 'cluster',
+                            'backup', 'high_availability', 'maintenance_window', 'auth_config']:
```
magodo: Is removing `create_mode` from `update_parameters` on purpose?

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -1049,6 +1128,22 @@ def format_item(self, item):
             result['auth_config']['tenant_id'] = item.auth_config.tenant_id
         else:
             result['auth_config'] = None
+        cluster_obj = None
+        if hasattr(item, 'cluster') and getattr(item, 'cluster') is not None:
+            cluster_obj = getattr(item, 'cluster')
+        elif hasattr(item, 'properties') and hasattr(item.properties, 'cluster') and getattr(item.properties, 'cluster') is not None:
+            cluster_obj = getattr(item.properties, 'cluster')
```
magodo: In which case the `item` maps to the top level object, and in which case it maps to the `properties`?

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -1049,6 +1128,22 @@ def format_item(self, item):
...
+        if cluster_obj is not None:
+            cluster_size = getattr(cluster_obj, 'cluster_size', None)
+            if cluster_size is None:
+                cluster_size = getattr(cluster_obj, 'clusterSize', None)
```
magodo: This is also suspicious, how come it is not consistent?

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -1049,6 +1128,22 @@ def format_item(self, item):
...
+
+            default_db = getattr(cluster_obj, 'default_database_name', None)
+            if default_db is None:
+                default_db = getattr(cluster_obj, 'defaultDatabaseName', None)
```
magodo: Same

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -800,6 +870,14 @@ def exec_module(self, **kwargs):
                 else:
                     self.update_parameters['storage'] = old_response['storage']
 
+                if self.update_parameters.get('cluster') is not None:
+                    for key in self.update_parameters['cluster'].keys():
+                        if (self.update_parameters['cluster'][key] is not None) and\
+                                (self.update_parameters['cluster'][key] != old_response['cluster'].get(key)):
+                            update_flag = True
+                        else:
+                            self.update_parameters['cluster'][key] = old_response['cluster'].get(key)
```
magodo: Is this necessary since they are already the same?

magodo (changes requested):
Thanks for this PR!
I've taken a look through and left some comments inline, but this is mostly looking good to me 👍

plugins/modules/azure_rm_postgresqlflexibleserver.py
zunyangc: Yes, because to support the elastic clusters, new SDK version is needed. The `begin_update` and `begin_create` has been removed and replaced with `begin_create_or_update` which uses the PUT request. 

https://pypi.org/project/azure-mgmt-postgresqlflexibleservers/#:~:text=Operation%20group%20ServersOperations%20added%20method%20begin_create_or_update

plugins/modules/azure_rm_postgresqlflexibleserver.py
zunyangc: Yes, because we switched from PATCH to PUT.

plugins/modules/azure_rm_postgresqlflexibleserver.py
zunyangc: Yes. `create_mode` field is only relevant when creating a server. Once the server exists, `create_mode` is not used.

plugins/modules/azure_rm_postgresqlflexibleserver.py
magodo: If the previous update is a PATCH operation, then after changing to PUT, the request parameter has the semantic changed, in which case you'll need to compose a *full* request body which contains all properties instead of the fileds get updated/patched.

plugins/modules/azure_rm_postgresqlflexibleserver.py
magodo: In that case, you'll always set the keys no matter they are changed or not.

magodo: In that case, you'll always set the keys no matter they are changed or not.

plugins/modules/azure_rm_postgresqlflexibleserver.py
```diff
@@ -1049,6 +1053,15 @@ def format_item(self, item):
...
             result['auth_config'] = None
+        cluster_obj = getattr(item, 'cluster', None)
+        if cluster_obj is None and hasattr(item, 'properties'):
+            cluster_obj = getattr(item.properties, 'cluster', None)
+        if cluster_obj is not None:
+            result['cluster']['cluster_size'] = getattr(cluster_obj, 'cluster_size', None)
+            result['cluster']['default_database_name'] = getattr(cluster_obj, 'default_database_name', None)
+        else:
+            result['cluster']['cluster_size'] = None
+            result['cluster']['default_database_name'] = None
```
magodo: Any reason not to follow the previous way of directly dereferencing fields through `item`, e.g. `item.cluster`?

magodo (changes requested):

plugins/modules/azure_rm_postgresqlflexibleserver.py
zunyangc: Thanks for letting me know! The pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=288898&view=results) passed. Thanks!

magodo (approved):

PR #2151, opened 2026-01-20
HCI improve calls

TiTi:
##### SUMMARY
Following #2077 
Prevents useless call if object is only Arc but not HCI

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
inventory azure_rm.py

##### ADDITIONAL INFORMATION
Splitted in 2 commits for readability (one with rename, one with the fix)

Happens if you use `include_hcivm_resource_groups` and you also have non HCI arc objects in targetted scope.
Without this PR, it causes requests to this handler, despite not being an `Microsoft.AzureStackHCI/virtualMachineInstances`:
`https://management.azure.com/subscriptions/<subscription_id>/resourceGroups/<rg>/providers/Microsoft.HybridCompute/machines/<arc name>/providers/Microsoft.AzureStackHCI/virtualMachineInstances?api-version=2024-01-01`

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -543,9 +543,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             self._enqueue_get(url=next_link, api_version=self._hybridcompute_api_version, handler=self._on_archcivm_page_response)
 
         for archcivm in response['value']:
-            url = '{0}/providers/Microsoft.AzureStackHCI/virtualMachineInstances'.format(archcivm['id'])
-            # Stack HCI instances look close enough to regular VMs that we can share the handler impl...
-            self._enqueue_get(url=url, api_version=self._stackhci_api_version, handler=self._on_vm_page_response, handler_args=dict(archcivm=archcivm))
+            if archcivm.get('kind') == 'HCI':
+                url = '{0}/providers/Microsoft.AzureStackHCI/virtualMachineInstances'.format(archcivm['id'])
+                # Stack HCI instances look close enough to regular VMs that we can share the handler impl...
+                self._enqueue_get(url=url, api_version=self._stackhci_api_version, handler=self._on_vm_page_response, handler_args=dict(archcivm=archcivm))
 
     def _on_vmss_page_response(self, response):
         next_link = response.get('nextLink')
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -371,14 +371,14 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
         url = url.format(subscriptionId=self._clientconfig.subscription_id, rg=rg)
         self._enqueue_get(url=url, api_version=self._hybridcompute_api_version, handler=self._on_arc_page_response)
 
-    def _enqueue_arcvm_list(self, rg='*'):
+    def _enqueue_archcivm_list(self, rg='*'):
         if not rg or rg == '*':
             url = '/subscriptions/{subscriptionId}/providers/Microsoft.HybridCompute/machines'
         else:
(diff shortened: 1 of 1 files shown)


plugins/inventory/azure_rm.py
```diff
@@ -536,16 +536,17 @@ def _on_arc_page_response(self, response):
...
         if next_link:
-            self._enqueue_get(url=next_link, api_version=self._hybridcompute_api_version, handler=self._on_arcvm_page_response)
+            self._enqueue_get(url=next_link, api_version=self._hybridcompute_api_version, handler=self._on_archcivm_page_response)
 
-        for arcvm in response['value']:
-            url = '{0}/providers/Microsoft.AzureStackHCI/virtualMachineInstances'.format(arcvm['id'])
-            # Stack HCI instances look close enough to regular VMs that we can share the handler impl...
-            self._enqueue_get(url=url, api_version=self._stackhci_api_version, handler=self._on_vm_page_response, handler_args=dict(arcvm=arcvm))
+        for archcivm in response['value']:
+            if archcivm.get('kind') == 'HCI':
```
zunyangc, line 546: Hi @TiTi, should this perhaps be checking for `if archcivm.get("kind") in ("HCI", "EPS"):` instead? 
Based on the current API versions in use, `self._hybridcompute_api_version = '2024-05-20-preview'` ([Reference](https://learn.microsoft.com/en-us/azure/templates/microsoft.hybridcompute/2024-05-20-preview/machines?pivots=deployment-language-arm-template#:~:text=%27AVS%27%0A%27AWS%27%0A%27EPS%27%0A%27GCP%27%0A%27HCI%27%0A%27SCVMM%27%0A%27VMware%27))

Let me know if I’ve misunderstood anything. Thanks!

zunyangc (changes requested):

plugins/inventory/azure_rm.py
TiTi: I have no such 'EPS' Kind in my ecosystem, despite having multiple HCI clusters. What does that stands for?
* AVS: Azure VMWare solution
* AWS: Amazon Web Services
* EPS: ?
* GCP: Google Cloud Platform
* HCI:  Azure Local
* SCVMM: System Center Virtual Machine Manager (Microsoft Windows)
* VMware: VMware by Broadcom, vSphere vCenter...

Best i could find are:
* Seqrite Endpoint Security (EPS) Cloud
* https://eps.millionminer.com/mobile-terms-and-conditions
* Oracle cloud appears to be Oracle Cloud Infrastructure (OCI)
* according to https://learn.microsoft.com/en-us/azure/azure-arc/overview it's Oracle but meh no idea

Anyway, outside HCI, all of those don't look like Microsoft.AzureStackHCI/virtualMachineInstances (Azure Local) resources to me ^^

zunyangc: Hi @TiTi, I couldn't find an explicit definition of EPS either, but all the HCI instances, I have access to return `kind: "EPS"`. 

From what I can tell, EPS stands for **Edge Platform Services**. They are actually a type `Microsoft.AzureStackHCI/edgeMachines`. 

Apologies for the confusion on my part. Happy to loop in @Klaas- as well if he has more context, but I agree, best to leave the condition as `kind=="HCI"` for now. 

Thanks!

TiTi: ah ok, this is probably [Azure Stack Hub](https://azure.microsoft.com/en-us/products/azure-stack/hub) then.
hmm yeah it's probably another REST endpoint that should be called then.

zunyangc:
Hi @magodo, pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=292519&view=results) passed. Ready to merge, thanks!

zunyangc (approved):

magodo (approved):

PR #2152, opened 2026-01-21
Fix 1222: Ensure container‑only operations use ARM management‑plane instead of requiring data‑plane access

### LINKED ISSUE #1222: azure_rm_storageblob  module requires data plane access even when only operating on management plane.
##### SUMMARY

The azure_rm_storageblob module is designed to manage both blob containers, and the blob files inside. The problem is that one of these operations, managing blob containers, takes place on the Azure RM management plane, and the other, managing the blobs inside, takes place on the data plane.

These have two totally different models of access, as well as networking, security, authentication, and permission schemes.

It should ideally be possible to access the Azure RM Management plane to manage the blob containers without needing to access the data plane, and possible to manage the blobs inside with no permissions to the management plane.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME

azure_rm_storageblob
##### ANSIBLE VERSION

```paste below
ansible [core 2.13.0]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/runner/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.9/site-packages/ansible
  ansible collection location = /home/runner/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.9.16 (main, May 31 2023, 12:21:58) [GCC 8.5.0 20210514 (Red Hat 8.5.0-18)]
  jinja version = 3.0.3
  libyaml = True
```

##### COLLECTION VERSION

```paste below
azure.azcollection 1.16.0
```

##### CONFIGURATION

```paste below

```
No output
##### OS / ENVIRONMENT

Ansible Controller 4.2.0

##### STEPS TO REPRODUCE

Create storage account with public network access set to disabled. Do not create private endpoint. Create blob container, failure.

```yaml

```

##### EXPECTED RESULTS

You should be able to create storage containers with no access to the data plane.

##### ACTUAL RESULTS

```paste below
HTTP response error: "This request is not authorized to perform this operation."
```
Note that this occurs after the same principal already created the storage account in same playbook run.

zunyangc:
##### SUMMARY
Fix #1222. This pull request updates `azure_rm_storageblob` to ensure container‑only operations (create, update, delete) correctly use the Azure Resource Manager (management plane) rather than accidentally invoking the data plane. Previously, the module incorrectly initialized or relied on a data‑plane client—even when no blob‑level operations were requested. 

Problem: If a user only wants to manage containers, they should not need data‑plane roles such as Storage Blob Data Contributor nor a working data‑plane network path.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
`plugins/modules/azure_rm_storageblob.py`

##### ADDITIONAL INFORMATION
- Added a new function `requires_data_plane_access`.
- Added `requires_data_plane_access` check in container-related functions.
- Only initialized data plane client if needed.

### DIFF
--- a/plugins/modules/azure_rm_storageblob.py
+++ b/plugins/modules/azure_rm_storageblob.py
@@ -404,9 +404,7 @@ class AzureRMStorageBlob(AzureRMModuleBase):
         elif self.state == 'absent':
             if not self.blob:
                 # Delete Container
-                # Ensure we can enforce "don't delete non-empty without force".
-                # If we don't already have a data-plane client and the caller didn't set force,
-                # try to initialize one just for the emptiness check.
+                # Initialize a data-plane client if caller force=false
                 if not self.blob_service_client and not self.force:
                     try:
                         self.blob_service_client = self.get_blob_service_client(
--- a/plugins/modules/azure_rm_storageblob.py
+++ b/plugins/modules/azure_rm_storageblob.py
@@ -268,6 +268,7 @@ import time
 try:
     from azure.storage.blob._models import BlobType, ContentSettings, StandardBlobTier
     from azure.core.exceptions import ResourceNotFoundError
+    from azure.mgmt.storage.models import BlobContainer
 except ImportError:
     # This is handled in azure_rm_common
     pass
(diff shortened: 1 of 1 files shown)


zunyangc:
Hi @magodo, pipeline test [result](https://dev.azure.com/azclitools/internal/_build/results?buildId=292429&view=results) passed. Ready to merge, thanks!

magodo (approved):

PR #2164, opened 2026-02-10
Fix crashes when processing VMs in failed or creating states

marespane:
##### SUMMARY
This PR fixes two critical crashes in the `azure_rm` inventory plugin that occur when processing Azure VMs in `Failed`, `Creating`, or other transient states.

In these states, the Azure SDK may return `None` for `instanceView` or omit the `networkProfile` key entirely from the VM model. The current code assumes these keys and objects always exist, leading to immediate inventory sync failures that prevent the plugin from processing the rest of the valid hosts.

Changes made:
1.  Added a check for `if self._instanceview` before attempting to access its attributes to prevent `AttributeError`.
2.  Switched to safe `.get()` calls when retrieving `networkInterfaces` from `networkProfile` to prevent `KeyError`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/inventory/azure_rm.py

##### ADDITIONAL INFORMATION
When an Azure subscription contains a VM in a broken state (e.g., "ProvisioningState/failed/NoRegisteredProviderFound"), the inventory plugin crashes with an unhandled exception, stopping the entire sync process.

The crash happens in two locations depending on the specific missing data:
1.  **AttributeError:** When `instanceView` is `None`, the code fails at `self._instanceview.get('computerName')`.
2.  **KeyError:** When the VM model lacks a `networkProfile`, the code fails when accessing `['networkInterfaces']`.

This fix ensures the plugin skips these missing attributes gracefully rather than crashing.

[WARNING]:  * Failed to parse /runner/inventory/azure_rm.yml with auto plugin: 'NoneType' object has no attribute 'get'
  File "/usr/share/ansible/collections/ansible_collections/azure/azcollection/plugins/inventory/azure_rm.py", line 852, in hostvars
    if self._instanceview.get('computerName'):
AttributeError: 'NoneType' object has no attribute 'get'

[WARNING]:  * Failed to parse /runner/inventory/azure_rm.yml with auto plugin: 'networkProfile'
  File "/usr/share/ansible/collections/ansible_collections/azure/azcollection/plugins/inventory/azure_rm.py", line 811, in hostvars
    nic_refs = vm_model['properties']['networkProfile']['networkInterfaces']
KeyError: 'networkProfile'

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -808,7 +808,7 @@ class AzureHost(object):
                                           api_version=self._inventory_client._compute_api_version,
                                           handler=self._on_instanceview_response)
 
-        nic_refs = vm_model['properties']['networkProfile']['networkInterfaces']
+        nic_refs = vm_model['properties'].get('networkProfile', dict()).get('networkInterfaces', [])
         for nic in nic_refs:
             # single-nic instances don't set primary, so figure it out...
             is_primary = nic.get('properties', {}).get('primary', len(nic_refs) == 1)
@@ -849,7 +849,7 @@ class AzureHost(object):
 
         createdAt = self._vm_model.get('systemData', {}).get('createdAt')  # hci specific
 
-        if self._instanceview.get('computerName'):
+        if self._instanceview and self._instanceview.get('computerName'):
             computer_name = self._instanceview['computerName']
         else:
             computer_name = self._vm_model['properties'].get('osProfile', {}).get('computerName')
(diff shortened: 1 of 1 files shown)


zunyangc:
Hi @marespane, thanks for opening up this PR.

Could you please provide steps to reproduce this error?

marespane:
hi @zunyangc, 
### Steps to Reproduce / Context
Reproducing this deterministically is difficult because it requires a Virtual Machine to be in a specific transient or failed state within the Azure backend.

**Environment Context:**
We operate a highly dynamic environment managing multiple Kubernetes clusters with high-churn Virtual Machine Scale Sets (VMSS). We frequently create and destroy ephemeral nodes.

**The Trigger Condition:**
The crash occurs when the Azure API returns a VM model that is incomplete due to a provisioning failure or a race condition during creation/deletion.

**Why the fix is needed:**
While enviroment like these are rare, the inventory plugin must be defensive and robust enough to handle these partial models without crashing. The proposed fix simply adds a guard clause to safely skip these incomplete records.

zunyangc:
Hi @marespane,

PR #2166 was recently merged into main and fixes the failing pipeline.
Please update your branch (rebase or merge main) so the PR validation picks up the latest changes.

Thanks!

marespane:
Hi @zunyangc,

I've pulled the changes from dev to my branch.

zunyangc:
Hi @magodo, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=298369&view=results) passed. Ready to merge, thanks!

magodo (approved):

PR #2206, opened 2026-03-30
feat: add support for feedback queue

mkanoor:
##### SUMMARY
ansible-rulebook now has an optional event persistence feature which allows the source plugin to know that the inflight event has been stored in a database and if there is an outage the event wont be lost. The Azure source plugins now have 2 extra parameters
 * feedback: bool : default (False)
 * feedback_timeout: int: default (120)

When feedback is set to True, ansible-rulebook will pass in an asyncio.Queue under the key "eda_feedback_queue". The source plugin can send in an event to ansible-rulebook and wait on the feedback Queue to get a response back after the rule engine has commited the event to disk if its needed. If its not needed the Event will be discarded.

This feature is only for in-flight events which are being processed. It is not meant to store every event in the Database.

https://redhat.atlassian.net/browse/AAP-66377

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
EDA Source Plugins
  Azure Event Hub
  Azure Service Bus

### DIFF
--- a/extensions/eda/plugins/event_source/azure_event_hub.py
+++ b/extensions/eda/plugins/event_source/azure_event_hub.py
@@ -9,8 +9,8 @@ import json
 import logging
 from typing import Any
 
-from azure.eventhub.aio import (EventHubConsumerClient, PartitionContext)
 from azure.eventhub import EventData
+from azure.eventhub.aio import EventHubConsumerClient, PartitionContext
 from azure.identity.aio import ClientSecretCredential
 
 DOCUMENTATION = r"""
@@ -56,6 +56,30 @@ options:
       - The name of the consumer group
     type: str
     default: "$Default"
+  feedback:
+    type: bool
+    default: false
+    description:
+      - Should the source plugin wait for feedback before
+        processing the next event from the Azure Event Hub
+        This flag allows ansible rulebook to pass in an asyncio
+        queue which is passed in the args['eda_feedback_queue']
+        The source plugin should wait for the response to come
+        back on this queue before it picks the next event from
+        Azure Event Hub.
+  feedback_timeout:
+    type: int
+    default: 120
+    description:
+      - Timeout in seconds to wait for feedback from the rule engine
+        before raising an exception. Only applies when feedback is enabled.
+  eda_feedback_queue:
+    description:
+      - Provided automatically by ansible-rulebook when the feedback parameter
+        is enabled, this parameter utilizes an asyncio.Queue. It allows the
+        system to wait for confirmation that an event has been safely persisted
+        in the database before removing it from the event bus. Users do not need
+        to provide a value for this manually.
 """
 
 EXAMPLES = r"""
@@ -79,6 +103,8 @@ REQUIRED_ARGS = [
     "azure_event_hub_name",
 ]
 
+DEFAULT_FEEDBACK_TIMEOUT = 120
+
 
 class AzureHubConsumer:
     """Azure Hub Consumer."""
@@ -95,6 +121,17 @@ class AzureHubConsumer:
 
         self.consumer_group = args.get("azure_consumer_group", "$Default")
         self.starting_position = int(args.get("azure_starting_position", "-1"))
+        self.feedback_timeout = int(args.get("feedback_timeout", DEFAULT_FEEDBACK_TIMEOUT))
+        self.feedback = args.get("feedback", False)
+        self.eda_feedback_queue = args.get("eda_feedback_queue")
+
+        if self.feedback and self.eda_feedback_queue is None:
+            msg = (
+                "feedback: true was set but no feedback queue was provided. "
+                "This requires a compatible version of ansible-rulebook that "
+                "supports the feedback mechanism."
+            )
+            raise ValueError(msg)
 
         self.credential = ClientSecretCredential(
             tenant_id=tenant_id,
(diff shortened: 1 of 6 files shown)


zunyangc:
Hi @mkanoor, thanks for this PR, we will look into this.

extensions/eda/plugins/event_source/azure_event_hub.py
```diff
@@ -56,6 +56,23 @@
...
     type: str
     default: "$Default"
+  feedback:
+    type: bool
+    default: false
+    description:
+      - Should the source plugin wait for feedback before
+        processing the next event from the Azure Event Hub
+        This flag allows ansible rulebook to pass in an asyncio
+        queue which is passed in the args['eda_feedback_queue']
```
p3ck, line 66: Shouldn't eda_feedback_queue be described in DOCUMENTATION like the rest of the possible args?

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -34,6 +35,23 @@
...
     type: bool
     default: true
+  feedback:
+    type: bool
+    default: false
+    description:
+      - Should the source plugin wait for feedback before
+        processing the next event from the Azure Event Hub
+        This flag allows ansible rulebook to pass in an asyncio
+        queue which is passed in the args['eda_feedback_queue']
```
p3ck, line 45: same comment.

mkanoor: @p3ck The eda_feedback_queue is an asyncio Queue that is created by ansible-rulebook it is not provided by the user. By using it in the args we can maintain backward compatibility with existing implementations.

extensions/eda/plugins/event_source/azure_event_hub.py
mkanoor: @p3ck The eda_feedback_queue is an asyncio Queue that is created by ansible-rulebook it is not provided by the user. By using it in the args we can maintain backward compatibility with existing implementations.

p3ck: Why is feedback documented properly then?  It doesn't make sense to me since you can't use feedback without defining the queue.

mkanoor: @p3ck feedback is an optional parameter that tells ansible-rulebook that this source plugin is expecting a feedback once the event has been persisted in the database to prevent loss during outage. When ansible-rulebook sees that a source needs feedback it creates an asyncio channel and passes that in via args['eda_feedback_queue'] so that the source plugin can wait for an acknowlegdement that the event is safe to deleted from the EventBus.

The same source plugin can work with and without feedback.

The function signature for any source plugin is **(input: asyncio.Queue, args: Dict)**
We could have changed it to **(input: asyncio.Queue, args: Dict, feedback_queue: asyncio.Queue)** that would have caused all source plugins to change. So a better backward compatible option was so keep the function signature the same and pass in the feedback queue via args with a special key called **eda_feedback_queue**.
The input queue is also created by ansible-rulebook similar to the feedback queue, these are not created by the users.

mkanoor: @p3ck I have added eda_feedback_queue in args stating that the user doesn't have to pass it in its automatically set when the user requests feedback.

extensions/eda/plugins/event_source/azure_service_bus.py
mkanoor: @p3ck I have added eda_feedback_queue in args stating that the user doesn't have to pass it in its automatically set when the user requests feedback.

p3ck (approved):
LGTM - Thanks

zunyangc:
Hi @magodo, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=307312&view=results) passed.

Ready to merge, thanks!
 
Unit test passed (PyTest output as follows):
````
~/azure$ python -m pytest tests/unit/event_source/test_azure_event_hub.py tests/unit/event_source/test_azure_service_bus.py -vvv

tests/unit/event_source/test_azure_event_hub.py::test_main_missing_required_args PASSED                                                              [  5%]
tests/unit/event_source/test_azure_event_hub.py::test_main_with_all_required_args PASSED                                                             [ 11%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_on_event_json PASSED                                                        [ 17%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_on_event_raw_string PASSED                                                  [ 23%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_on_event_unicode_error PASSED                                               [ 29%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_on_event_no_event PASSED                                                    [ 35%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_initialization PASSED                                                       [ 41%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_defaults PASSED                                                             [ 47%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_start_receiving PASSED                                                      [ 52%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_feedback_enabled_without_queue PASSED                                       [ 58%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_feedback_enabled_with_queue PASSED                                          [ 64%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_feedback_timeout PASSED                                                     [ 70%]
tests/unit/event_source/test_azure_event_hub.py::test_azure_hub_consumer_feedback_defaults PASSED                                                    [ 76%]
tests/unit/event_source/test_azure_service_bus.py::test_receive_events PASSED                                                                        [ 82%]
tests/unit/event_source/test_azure_service_bus.py::test_receive_events_feedback_enabled_without_queue PASSED                                         [ 88%]
tests/unit/event_source/test_azure_service_bus.py::test_receive_events_feedback_enabled_with_queue PASSED                                            [ 94%]
tests/unit/event_source/test_azure_service_bus.py::test_receive_events_feedback_timeout PASSED                                                       [100%]

==================================================================== 17 passed in 0.79s ====================================================================
````

PR #2112, opened 2025-11-18
Add suppression for new 2.21 sanity check

p3ck:
##### SUMMARY
Disable 2.21 sanity check on these modules for now

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_automationaccount_info.py
plugins/modules/azure_rm_cognitivesearch_info.py
plugins/modules/azure_rm_imagesku_info.py
plugins/modules/azure_rm_keyvault_info.py
plugins/modules/azure_rm_servicebussaspolicy.py
plugins/modules/azure_rm_tags_info.py
plugins/modules/azure_rm_vmsku_info.py

PR #2210, opened 2026-04-07
Fix #2126: Fix reserved parameters and return keys

### LINKED ISSUE #2126: Fix for reserved parameters and return keys
##### SUMMARY
Latest ansible-core and sanity tests will warn or fail when reserved keys or parameters are used.

ie: name, tags

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
plugins/modules/azure_rm_automationaccount_info.py
plugins/modules/azure_rm_cognitivesearch_info.py
plugins/modules/azure_rm_imagesku_info.py
plugins/modules/azure_rm_keyvault_info.py
plugins/modules/azure_rm_servicebussaspolicy.py
plugins/modules/azure_rm_tags_info.py
plugins/modules/azure_rm_vmsku_info.py

##### ANSIBLE VERSION
ansible-2.20

##### COLLECTION VERSION
3.12.0

##### Possible solution

The vmware collection has provided a backwards compatible solution

https://github.com/ansible-collections/vmware.vmware/pull/286/files

zunyangc:
##### SUMMARY

**FUTURE ACTION NEEDED:**
- Remove [`ignore-2.21.txt line 1-7`](https://github.com/ansible-collections/azure/blob/023679ed967087cdce89e84203c80e7d55f2adef/tests/sanity/ignore-2.21.txt#L1-L7) after 2027-05-01.
- Remove old keys reference in module code and RETURN docstring.

Fix #2126, related to #2112

﻿Renames reserved Python built-in return value keys (keys, values) that conflict with Jinja2 dot-notation access. ﻿
These `validate-modules:bad-return-value-key` sanity errors are reported starting with `ansible-core 2.21` (currently in `devel`). They were temporarily suppressed via `tests/sanity/ignore-2.21.txt`. The 7 ignore entries can be removed after the removal date of `2027-05-01`.

Each affected module now: 
1. Adds a new non-reserved return key as the primary documented key
2. Keeps the old key populated (aliased to the new key) for backward compatibility
3. Marks the old key as deprecated in the RETURN docstring with a removal date of `2027-05-01`
4. Emits a runtime deprecation warning via `module.deprecate()`

| Module | Old Key | New Key |
| ------------- | ------------- | ------------- |
| azure_rm_automationaccount_info | `keys` | `account_keys` |
| azure_rm_cognitivesearch_info | `keys` | `search_keys` |
| azure_rm_keyvault_info | `keys` (in `permissions`) | `key_permissions` |
| azure_rm_servicebussaspolicy | `keys` | `sas_keys` |
| azure_rm_imagesku_info | `values` (in `restrictions`) | `restriction_values` |
| azure_rm_vmsku_info | `values` (in `restrictions`) | `restriction_values` |
| azure_rm_tags_info | `values` (in `tag_details`) | `tag_values` |

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
- azure_rm_automationaccount_info
- azure_rm_cognitivesearch_info
- azure_rm_keyvault_info
- azure_rm_servicebussaspolicy
- azure_rm_imagesku_info
- azure_rm_vmsku_info
- azure_rm_tags_info

##### ADDITIONAL INFORMATION
Sample deprecation warning:
````
[DEPRECATION WARNING]: The 'keys' return key in 'permissions' is deprecated. Use 'key_permissions' instead. This feature will be removed from collection 'azure.azcollection' in a release after 2027-05-01.
````

### DIFF
--- a/plugins/modules/azure_rm_automationaccount_info.py
+++ b/plugins/modules/azure_rm_automationaccount_info.py
@@ -124,9 +124,35 @@ automation_accounts:
             type: str
             returned: always
             sample: ok
+        account_keys:
+            description:
+                - Resource keys.
+            type: complex
+            returned: always
+            contains:
+                key_name:
+                    description:
+                        - Name of the key.
+                    type: str
+                    returned: always
+                    sample: Primary
+                permissions:
+                    description:
+                        - Permission of the key.
+                    type: str
+                    returned: always
+                    sample: Full
+                value:
+                    description:
+                        - Value of the key.
+                    type: str
+                    returned: always
+                    sample: "MbepKTO6IyGwml0GaKBkKN"
         keys:
             description:
                 - Resource keys.
+                - This return value has been deprecated and will be removed in a release after
+                  2027-05-01. Use C(account_keys) instead.
             type: complex
             returned: always
             contains:
@@ -285,6 +311,13 @@ class AzureRMAutomationAccountInfo(AzureRMModuleBase):
         for key in list(self.module_arg_spec):
             setattr(self, key, kwargs[key])
 
+        if self.list_keys:
+            self.module.deprecate(
+                "The 'keys' return key is deprecated. Use 'account_keys' instead.",
+                date="2027-05-01",
+                collection_name="azure.azcollection",
+            )
+
         if self.resource_group and self.name:
             accounts = [self.get()]
         elif self.resource_group:
@@ -305,7 +338,8 @@ class AzureRMAutomationAccountInfo(AzureRMModuleBase):
         if self.list_usages:
             result['usages'] = self.get_usages(id_dict['resource_group'], account.name)
         if self.list_keys:
-            result['keys'] = self.list_account_keys(id_dict['resource_group'], account.name)
+            result['account_keys'] = self.list_account_keys(id_dict['resource_group'], account.name)
+            result['keys'] = result['account_keys']
         return result
 
     def get(self):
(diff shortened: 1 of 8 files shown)


Klaas-:
is this something that should be addressed in the azure python sdk as well?

zunyangc:
Hi @p3ck, it would be great to have your input on this PR when you have time, thanks!
- If we keep `ignore-2.21.txt` going forward, do you see any potential issues with that approach?
- Also, is this behavior specific to `ansible-core==2.21` (`devel`)? In #2126 , it appears a similar issue was also observed with `ansible-core==2.20` on your end. Could you help clarify?
- There is also `ansible-bad-function: Call run_command instead of subprocess.run` which is related to Pylint, currently being suppressed.

zunyangc:
> is this something that should be addressed in the azure python sdk as well?

Hi @Klaas-, from what I can tell, this seems to be more of an Ansible/Jinja2-layer issue. The SDK property names like keys and values are valid Python - the issue that when these are used as Ansible return keys, Jinja2 dot-notation (e.g. `result.keys`) can be resolve to Python `dict.keys()` method instead of the actual data. That's what the `validate-modules:bad-return-value-key` sanity check in ansible-core (devel) is flagging.

So I think the fix is appropriately scoped at the collection layer, but happy to hear other thoughts on this. Ideally, we want to avoid changing existing behavior.

Reference: https://github.com/ansible/ansible/pull/86079

p3ck (approved):
This looks good to me.  I don't see any other option than what you have done here since we don't know what variables the user is "using".

PR #2223, opened 2026-04-17
Fix #1223: `subscription_id` module parameter ignored in resource ID construction

### LINKED ISSUE #1223: The `subscription_id` Module Parameter Is Used Contradictively

zunyangc:
##### SUMMARY
Fix #1223 

**Troubleshooting**
Problem: 
The `env` and `credential_file` auth sources inside `AzureRMAuth._get_credentials()` ignored the user-specified 
`subscription_id` module parameter. When a user passed `subscription_id` to target a specific subscription but
authenticated via environment variables or credential file, the `subscription_id` was silently dropped — either causing
auth failure (if `AZURE_SUBSCRIPTION_ID` wasn't set) or using the wrong subscription (if it was set to a different
value). This also affected the auto auth source, which cascades through env and credential_file paths.

Root Cause: 
`_get_env_credentials()` and `_get_profile()` only read from their own sources (env vars / credential file) and had no way to accept the user-specified `subscription_id` passed into `_get_credentials()`. Meanwhile, `_get_msi_credentials()` and `_get_azure_cli_credentials()` already honored it correctly — the behavior was inconsistent across auth sources

 **Fix**
1. Inside `AzureRMAuth` - `_get_env_credentials()` and _get_profile() now accept an optional `subscription_id` parameter and overlay it onto their source-resolved credentials. All call sites in `_get_credentials()` pass `subscription_id` through.
2. Simplify `get_mgmt_svc_client` to use `self.subscription_id` directly instead of duplicating the override logic, since upstream resolution is now correct.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/azure_rm_common.py

##### ADDITIONAL INFORMATION
- When `subscription_id` is not provided: behavior is unchanged - auth sources resolve it from their own defaults
- When `subscription_id` matches auth subscription: behavior is unchanged
- When `subscription_id` differs from auth subscription: all auth paths now correctly honor the user-specified value (previously ignored by env/credential_file/auto)

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1620,7 +1620,7 @@ class AzureRMAuth(object):
         if self.credentials.get('subscription_id', None) is None and not self.is_ad_resource:
             self.fail("Credentials did not include a subscription_id value.")
         self.log("setting default_subscription_id")
-        # The default subscription resolved during auth setup. 
+        # The default subscription resolved during auth setup.
         # This is NOT used for authentication itself (Azure auth doesn't require a subscription ID).
         # It serves as the fallback when the user doesn't explicitly provide a subscription_id module parameter. Sources vary by auth method:
         #   - CLI: active subscription from `az account show`
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -890,8 +890,9 @@ class AzureRMModuleBase(object):
         if not base_url.endswith("/"):
             base_url += "/"
 
-        # Use the subscription_id property which prefers the user-provided module param over the default subscription resolved during auth setup.
-        mgmt_subscription_id = self.subscription_id
+        mgmt_subscription_id = self.azure_auth.subscription_id
+        if self.module.params.get('subscription_id'):
+            mgmt_subscription_id = self.module.params.get('subscription_id')
 
         # Some management clients do not take a subscription ID as parameters.
         if suppress_subscription_id:
@@ -993,9 +994,7 @@ class AzureRMModuleBase(object):
 
     @property
     def subscription_id(self):
-        # Prefer the user-provided subscription_id module param over the default subscription resolved during auth setup.
-        # This ensures resource ID construction and SDK clients consistently target the subscription the user explicitly specified.
-        return self.module.params.get('subscription_id') or self.azure_auth.default_subscription_id
+        return self.azure_auth.subscription_id
 
     @property
     def storage_client(self):
@@ -1619,15 +1618,8 @@ class AzureRMAuth(object):
 
         if self.credentials.get('subscription_id', None) is None and not self.is_ad_resource:
             self.fail("Credentials did not include a subscription_id value.")
-        self.log("setting default_subscription_id")
-        # The default subscription resolved during auth setup.
-        # This is NOT used for authentication itself (Azure auth doesn't require a subscription ID).
-        # It serves as the fallback when the user doesn't explicitly provide a subscription_id module parameter. Sources vary by auth method:
-        #   - CLI: active subscription from `az account show`
-        #   - MSI: first subscription the managed identity can access
-        #   - Env: AZURE_SUBSCRIPTION_ID environment variable
-        #   - Credential file: subscription_id from ~/.azure/credentials profile
-        self.default_subscription_id = self.credentials.get('subscription_id')
+        self.log("setting subscription_id")
+        self.subscription_id = self.credentials.get('subscription_id')
 
         # get authentication authority
         # for adfs, user could pass in authority or not.
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -450,13 +450,7 @@ class AzureRMModuleBase(object):
         # self.debug = self.module.params.get('debug')
 
         # delegate auth to AzureRMAuth class (shared with all plugin types)
-        # Resolve env var fallbacks before passing params, so AzureRMAuth
-        # receives the user's full intent (module param > env var > None).
-        auth_params = dict(self.module.params)
-        for attr, env_var in AZURE_CREDENTIAL_ENV_MAPPING.items():
-            if auth_params.get(attr) is None:
-                auth_params[attr] = os.environ.get(env_var, None)
-        self.azure_auth = AzureRMAuth(fail_impl=self.fail, is_ad_resource=is_ad_resource, **auth_params)
+        self.azure_auth = AzureRMAuth(fail_impl=self.fail, is_ad_resource=is_ad_resource, **self.module.params)
 
         # common parameter validation
         if self.module.params.get('tags'):
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -890,9 +890,8 @@ class AzureRMModuleBase(object):
         if not base_url.endswith("/"):
             base_url += "/"
 
-        mgmt_subscription_id = self.azure_auth.subscription_id
-        if self.module.params.get('subscription_id'):
-            mgmt_subscription_id = self.module.params.get('subscription_id')
+        # Use the subscription_id property which prefers the user-provided module param over the default subscription resolved during auth setup.
+        mgmt_subscription_id = self.subscription_id
 
         # Some management clients do not take a subscription ID as parameters.
         if suppress_subscription_id:
@@ -994,8 +993,9 @@ class AzureRMModuleBase(object):
 
     @property
     def subscription_id(self):
-        # Prefer user-provided subscription_id module param over auth subscription
-        return self.module.params.get('subscription_id') or self.azure_auth.subscription_id
+        # Prefer the user-provided subscription_id module param over the default subscription resolved during auth setup.
+        # This ensures resource ID construction and SDK clients consistently target the subscription the user explicitly specified.
+        return self.module.params.get('subscription_id') or self.azure_auth.default_subscription_id
 
     @property
     def storage_client(self):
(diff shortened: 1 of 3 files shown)


zunyangc:
Hi @magodo, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=310178&view=results) passed. Ready to merge, thanks!

plugins/module_utils/azure_rm_common.py
```diff
@@ -890,9 +890,8 @@ def get_mgmt_svc_client(self, client_type, base_url=None, api_version=None, supp
         if not base_url.endswith("/"):
             base_url += "/"
 
-        mgmt_subscription_id = self.azure_auth.subscription_id
-        if self.module.params.get('subscription_id'):
-            mgmt_subscription_id = self.module.params.get('subscription_id')
+        # Use the subscription_id property which prefers the user-provided module param over the default subscription resolved during auth setup.
```
magodo: This comment is not needed as the usage here is clear. We don't have to mention the reason about the change in the comment, just enclose the conversation in the PR.

plugins/module_utils/azure_rm_common.py
```diff
@@ -994,7 +993,9 @@ def _cloud_environment(self):
 
     @property
     def subscription_id(self):
-        return self.azure_auth.subscription_id
+        # Prefer the user-provided subscription_id module param over the default subscription resolved during auth setup.
+        # This ensures resource ID construction and SDK clients consistently target the subscription the user explicitly specified.
+        return self.module.params.get('subscription_id') or self.azure_auth.default_subscription_id
```
magodo: I prefer to keep the original logic to simply make the `subscription_id` an alias property to the `azure_auth.subscription_id`, but fix the inconsistency in the subscription resolving inside the problematic auth modes.

plugins/module_utils/azure_rm_common.py
```diff
@@ -1618,8 +1619,15 @@ def __init__(self, auth_source=None, profile=None, subscription_id=None, client_
...
-        self.log("setting subscription_id")
-        self.subscription_id = self.credentials.get('subscription_id')
+        self.log("setting default_subscription_id")
+        # The default subscription resolved during auth setup.
+        # This is NOT used for authentication itself (Azure auth doesn't require a subscription ID).
+        # It serves as the fallback when the user doesn't explicitly provide a subscription_id module parameter. Sources vary by auth method:
+        #   - CLI: active subscription from `az account show`
+        #   - MSI: first subscription the managed identity can access
+        #   - Env: AZURE_SUBSCRIPTION_ID environment variable
+        #   - Credential file: subscription_id from ~/.azure/credentials profile
```
magodo: Again these comments are unnecessary

plugins/module_utils/azure_rm_common.py
```diff
@@ -1618,8 +1619,15 @@ def __init__(self, auth_source=None, profile=None, subscription_id=None, client_
...
+        self.default_subscription_id = self.credentials.get('subscription_id')
```
magodo: If we can fix the subscription resolution in the auth part, this attribute can be just renamed back to `subscription_id`.

magodo (changes requested):
I took a look at the code and feel the issue actually is derived from the inconsistent implementation inside `AzureRMAuth`.

It is firstly instantiated at: https://github.com/ansible-collections/azure/blob/a8e69e9efce6665ed4e3575688148a57566bb8c1/plugins/module_utils/azure_rm_common.py#L456-L457

Here the `subscription_id` is passed from the params, if specified. The first improvement we can do here might be to take the env mapping into consideration right before we pass the params: if one param is not set, then fallback to its corresponding env var. By doing this, we don't need to care about env var inside of `AzureRMAuth`.

The code continue flows to: https://github.com/ansible-collections/azure/blob/a8e69e9efce6665ed4e3575688148a57566bb8c1/plugins/module_utils/azure_rm_common.py#L1571-L1588, where the `credentials` is constructed, again here the subscription id is properly passed in.

Inside the `_get_credentials()`, there are a couple of auth source going on:

- `msi`: It propery honors the specified subscription id (though if we handle the env var before, it can be even further simplified to get rid of https://github.com/ansible-collections/azure/blob/a8e69e9efce6665ed4e3575688148a57566bb8c1/plugins/module_utils/azure_rm_common.py#L1787)
- `cli`: It also behaves correctly
- `env`: This one can be deprecated if we want to honor the env vars as a fallback of the parameters, whilst I'd suggest to deprecate it in another PR. The problem of this one is it totally ignores the params users specified, this might be on purpose but questionable and confusing
- `credential_file`: This one is even more confusing as it reuses the `_get_profile()` as the `env` source does. Hence it ignores the user specified param as well.
- *auto*: This has the same problem for a couple of cases.

My suggestion is to tidy them up to behave consistently, follow the same resolution order for subscription_id, and the many other attributes.

plugins/module_utils/azure_rm_common.py
```diff
@@ -450,7 +450,13 @@ def __init__(self, derived_arg_spec, bypass_checks=False, no_log=False,
         # self.debug = self.module.params.get('debug')
 
         # delegate auth to AzureRMAuth class (shared with all plugin types)
-        self.azure_auth = AzureRMAuth(fail_impl=self.fail, is_ad_resource=is_ad_resource, **self.module.params)
+        # Resolve env var fallbacks before passing params, so AzureRMAuth
+        # receives the user's full intent (module param > env var > None).
+        auth_params = dict(self.module.params)
+        for attr, env_var in AZURE_CREDENTIAL_ENV_MAPPING.items():
+            if auth_params.get(attr) is None:
+                auth_params[attr] = os.environ.get(env_var, None)
```
magodo: We shall not touch this part to avoid any potential breaking change (though this logic is more proper).

magodo (changes requested):

magodo (approved):

PR #2237, opened 2026-05-06
feat(inventory): log filter expression when azure_rm host is excluded

stevefulme1:
## Summary

- Adds verbose logging (`display.v`) to the `azure_rm` inventory plugin's `_filter_host()` method
- When a host is excluded by `exclude_host_filters`, `default_host_filters`, or `include_host_filters`, the specific filter expression that matched is now logged at verbosity level 1 (`-v`)
- Messages only appear with `-v` flag — no noise at default verbosity

## Problem

When hosts are excluded by filter rules, there is no log output explaining which filter caused the exclusion. This makes it difficult to troubleshoot why hosts disappear from inventory, especially when using `overwrite: true` in AAP inventory syncs where silent exclusion can cause destructive host deletion.

## Changes

- `_filter_host()`: Log the matching filter expression at `display.v()` when a condition evaluates to `True`
- `_populate()`: Log at `display.v()` when a host is skipped due to exclude or include filters, with context about which filter type caused it

## Example output (with `-v`)

```
azure_rm inventory: host 'my-vm-01' matched filter: powerstate != 'running'
azure_rm inventory: host 'my-vm-01' excluded by exclude_host_filters/default_host_filters
```

## Resolves

- [AAPRFE-2898](https://redhat.atlassian.net/browse/AAPRFE-2898) — Improve logging around azure_rm inventory plugin to log the filter that excluded a host

### DIFF
--- a/plugins/inventory/azure_rm.py
+++ b/plugins/inventory/azure_rm.py
@@ -434,8 +434,10 @@ class InventoryModule(BaseInventoryPlugin, Constructable, Cacheable):
             display.vvvv(f"Populate {inventory_hostname}")
 
             if self._filter_exclude_host(inventory_hostname, hostvars):
+                display.v(f"azure_rm inventory: host '{inventory_hostname}' excluded by exclude_host_filters/default_host_filters")
                 continue
             if not self._filter_include_host(inventory_hostname, hostvars):
+                display.v(f"azure_rm inventory: host '{inventory_hostname}' excluded by include_host_filters (no include rule matched)")
                 continue
 
             try:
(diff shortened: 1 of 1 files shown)


p3ck (approved):
LGTM - Thanks!

zunyangc (approved):
LGTM, thanks

TiTi:
hey, can we put that with at least .vv verbosity level please?
it could generate huge output on some inventories

TiTi:
please note that it will output at least 1 line for each machine of the inventory even if you use --limit on specific machines

PR #2263, opened 2026-05-29
Feature #2257: List active certificates in `azure_rm_keyvaultcertificate_info` when name is omitted

### LINKED ISSUE #2257: Add Feature to list active Certificates with azure_rm_keyvaultcertificate_info
##### SUMMARY
Currently, the `azure_rm_keyvaultcertificate_info` module requires the `name` parameter to fetch active certificates. If the `name` parameter is omitted, the module silently exits without returning active certificates (it currently only processes `show_deleted_certificate` in the `else` block). 

This feature request proposes making the `name` parameter optional (`required=False`), so that omitting it will return a list of all active certificates in the Key Vault. This can be achieved by leveraging the underlying `list_properties_of_certificates()` method from the `azure-keyvault-certificates` SDK.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_keyvaultcertificate_info

##### ADDITIONAL INFORMATION
Allowing the module to list all certificates when `name` is omitted would align its behavior with standard Ansible info modules (which generally return a list of all resources when a specific identifier is omitted) and eliminate the need for CLI workarounds.

```yaml
---
- name: Get all certificates from a Key Vault
  hosts: localhost
  tasks:
    - name: List all active Key Vault certificates (Desired Behavior)
      azure.azcollection.azure_rm_keyvaultcertificate_info:
        vault_uri: "https://my-key-vault.vault.azure.net"
        # 'name' is intentionally omitted to fetch all certificates
      register: all_certs

    - name: Display certificate expirations
      ansible.builtin.debug:
        msg: "Certificate {{ item.properties.name }} expires on {{ item.properties.attributes.expires }}"
      loop: "{{ all_certs.certificates }}"

### PR DESCRIPTION
**Pull request by zunyangc**
##### SUMMARY
Fix #2257.

`azure_rm_keyvaultcertificate_info` already documented that omitting `name` would list all certificates in the vault, but the implementation only honoured that promise when `show_deleted_certificate=true`. With the default `show_deleted_certificate=false`, omitting `name` silently returned no `certificates` key at all — a doc&code mismatch that this PR closes.

##### ISSUE TYPE
- Bugfix Pull Request
- Feature Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_keyvaultcertificate_info.py
tests/integration/targets/azure_rm_keyvaultcertificate/tasks/main.yml

### DIFF
--- a/plugins/modules/azure_rm_keyvaultcertificate_info.py
+++ b/plugins/modules/azure_rm_keyvaultcertificate_info.py
@@ -23,7 +23,9 @@ options:
         type: str
     name:
         description:
-            - Certificate name. If not set, will list all certificates in vault_uri.
+            - Certificate name.
+            - If not set, when I(show_deleted_certificate=false) (the default) the module returns the list of all active certificates in I(vault_uri).
+            - If not set and I(show_deleted_certificate=true), the module returns the list of all deleted certificates in I(vault_uri).
         type: str
     version:
         description:
@@ -34,6 +36,14 @@ options:
             - Set to I(show_delete_certificate=true) to show deleted certificates. Set to I(show_deleted_certificate=false) to show not deleted certificates.
         type: bool
         default: false
+    include_pending:
+        description:
+            - When listing all active certificates (I(name) omitted and I(show_deleted_certificate=false)),
+              include certificates that are not yet completely provisioned.
+            - Only honoured by the Key Vault REST API v7.0 and above. If omitted, Key Vault treats this as C(false).
+            - Has no effect when I(name) is specified or when I(show_deleted_certificate=true).
+        type: bool
+        version_added: "3.20.0"
     tags:
         description:
             - Limit results by providing a list of tags. Format tags as 'key' or 'key:value'.
@@ -71,6 +81,15 @@ EXAMPLES = '''
   azure_rm_keyvaultcertificate_info:
     vault_uri: "https://myVault.vault.azure.net"
     show_deleted_certificate: true
+
+- name: List all active certificates in the key vault
+  azure.azcollection.azure_rm_keyvaultcertificate_info:
+    vault_uri: "https://myVault.vault.azure.net"
+
+- name: List all active certificates including pending ones
+  azure.azcollection.azure_rm_keyvaultcertificate_info:
+    vault_uri: "https://myVault.vault.azure.net"
+    include_pending: true
 '''
 
 RETURN = '''
@@ -487,12 +506,14 @@ class AzureRMKeyVaultCertificateInfo(AzureRMModuleBase):
                                     name=dict(type='str'),
                                     vault_uri=dict(type='str', required=True),
                                     show_deleted_certificate=dict(type='bool', default=False),
+                                    include_pending=dict(type='bool'),
                                     tags=dict(type='list', elements='str'))
 
         self.vault_uri = None
         self.name = None
         self.version = None
         self.show_deleted_certificate = False
+        self.include_pending = None
         self.tags = None
 
         self.results = dict(changed=False)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by zunyangc** [comment]
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=319763&view=results) passed. Ready to merge, thanks!

File: plugins/modules/azure_rm_keyvaultcertificate_info.py [hunk 1]
```diff
@@ -621,6 +644,43 @@ def list_deleted_certificates(self):
...
+            response = self._client.list_properties_of_certificates(**list_kwargs)
+            self.log("Response : {0}".format(response))
+
+            if response:
+                for item in response:
+                    # Skip confirmed-disabled certs
+                    if item.enabled is False:
+                        continue
+                    try:
+                        bundle = self._client.get_certificate(certificate_name=item.name)
```
Vieran: Any reason here to get detail of every certificates? Usually `list` only give a brief of the items and if the user want detail, they could call `get_certificate` after get the list.

tests/integration/targets/azure_rm_keyvaultcertificate/tasks/main.yml
```diff
@@ -166,6 +166,48 @@
...
+          - list_active_facts['certificates'] is iterable
+          # imported "-secondary" cert is enabled by default and should be present;
+          # the original "cert{{ rpfx }}" is disabled and is intentionally skipped.
+          - list_active_facts['certificates'] | length >= 1
+          - "'cert' + rpfx + '-secondary' in (list_active_facts['certificates'] | map(attribute='name') | list)"
+          - "'cert' + rpfx not in (list_active_facts['certificates'] | map(attribute='name') | list)"
+          - list_active_facts['certificates'][0]['policy'] is defined
+          - list_active_facts['certificates'][0]['properties']['attributes'] is defined
+
+    - name: List all active keyvault certificates with include_pending=true
```
Vieran: Is this test really construct a keyvault with `pending` certificate?

tests/integration/targets/azure_rm_keyvaultcertificate/tasks/main.yml
```diff
@@ -166,6 +166,48 @@
...
+      ansible.builtin.assert:
+        that:
+          - list_active_pending_facts['certificates'] is defined
+          - list_active_pending_facts['certificates'] is iterable
+
+    - name: List active certificates with a non-matching tag filter
+      azure.azcollection.azure_rm_keyvaultcertificate_info:
+        vault_uri: https://vault{{ rpfx }}.vault.azure.net
+        tags:
+          - this_tag_does_not_exist
```
Vieran: Is this a `non-matching` tag or not a tag?

tests/integration/targets/azure_rm_keyvaultcertificate/tasks/main.yml
```diff
@@ -166,6 +166,48 @@
...
+        include_pending: true
+      register: list_active_pending_facts
+
+    - name: Assert include_pending listing succeeds
+      ansible.builtin.assert:
+        that:
+          - list_active_pending_facts['certificates'] is defined
+          - list_active_pending_facts['certificates'] is iterable
+
+    - name: List active certificates with a non-matching tag filter
```
Vieran: Maybe add a positive test case to test that the code could correctly find those matching tag certificates?

Vieran (changes requested):

tests/integration/targets/azure_rm_keyvaultcertificate/tasks/main.yml
zunyangc: As for now, I couldn't find any way that could leave the cert in a pending state (might need a non-self-assigned issuer). Hence, removing this test block as no-op. 

Thanks!

Vieran (approved):

PR #2276, opened 2026-06-09
Fix #2274: Honour `AZURE_TENANT_ID` and caller-supplied `subscription_id` in `azure_rm` workload-identity auth

### LINKED ISSUE #2274: azure_rm: workload identity auth reads AZURE_TENANT but not AZURE_TENANT_ID
##### SUMMARY

When authenticating with Microsoft Entra Workload Identity (for example on AKS with the workload-identity webhook), the `azure_rm` OIDC path resolves the tenant only from `AZURE_TENANT`. The webhook, and the `azure-identity` SDK, use `AZURE_TENANT_ID`.

The collection already reads `AZURE_CLIENT_ID` and `AZURE_FEDERATED_TOKEN_FILE`, both of which match what the webhook injects, so the client id and the federated token resolve fine. Only the tenant is out of step, so the sync fails to authenticate until the user manually re-exports `AZURE_TENANT=$AZURE_TENANT_ID` in the pod.

##### COMPONENT NAME

`azure_rm` inventory plugin (auth in `plugins/module_utils/azure_rm_common.py`)

##### DETAILS

The env mapping reads tenant from `AZURE_TENANT` ([azure_rm_common.py#L57-L73](https://github.com/ansible-collections/azure/blob/v3.18.0/plugins/module_utils/azure_rm_common.py#L57-L73)):

```python
AZURE_CREDENTIAL_ENV_MAPPING = dict(
    profile='AZURE_PROFILE',
    subscription_id='AZURE_SUBSCRIPTION_ID',
    client_id='AZURE_CLIENT_ID',
    secret='AZURE_SECRET',
    tenant='AZURE_TENANT',
    ...
    oidc_token='AZURE_FEDERATED_TOKEN',
    oidc_token_file_path='AZURE_FEDERATED_TOKEN_FILE'
)
```

The OIDC token-file branch ([azure_rm_common.py#L1663-L1679](https://github.com/ansible-collections/azure/blob/v3.18.0/plugins/module_utils/azure_rm_common.py#L1663-L1679)) builds the credential with that tenant:

```python
elif self.credentials.get('client_id') is not None and \
        self.credentials.get('tenant') is not None and \
        self.credentials.get('oidc_token_file_path') is not None:
    token_file = self.credentials['oidc_token_file_path']

    if not os.path.exists(token_file):
        self.fail(f"The specified OIDC token file does not exist: {token_file}")

    def _load_assertion():
        with open(token_file) as f:
            return f.read()

    self.azure_credential_track2 = ClientAssertionCredential(tenant_id=self.credentials['tenant'],
                                                             client_id=self.credentials['client_id'],
                                                             func=_load_assertion,
                                                             authority=self._adfs_authority_url,
                                                             disable_instance_discovery=self._disable_instance_discovery)
```

The AKS workload-identity webhook injects `AZURE_CLIENT_ID`, `AZURE_TENANT_ID`, `AZURE_FEDERATED_TOKEN_FILE` and `AZURE_AUTHORITY_HOST`. So `client_id` and the token file line up with the names the plugin reads, but the tenant does not (`AZURE_TENANT_ID` injected vs `AZURE_TENANT` read).

##### SUGGESTED FIX

Fall back to `AZURE_TENANT_ID` when `AZURE_TENANT` is unset, so workload identity works without manually re-exporting the variable. Happy to send a PR with a changelog fragment if that approach sounds reasonable.

A related rough edge in the same area: `_get_env_credentials` returns `None` when `AZURE_SUBSCRIPTION_ID` is unset, before a plugin-supplied `subscription_id` is applied, so a file `subscription_id` is not honored under workload identity. Probably worth a look too, but the tenant fallback is the main ask.

##### VERSION

azure.azcollection 3.18.0 (line references above are to the `v3.18.0` tag; `dev` still maps `tenant` to `AZURE_TENANT`).

zunyangc:
##### SUMMARY
Fix #2274.

The `azure_rm` modules and the `azure_rm` inventory plugin did not pick up the environment variables injected by the AKS workload-identity webhook (and used by the `azure-identity` SDK), so OIDC auth failed out-of-the-box on AKS workloads until the user manually re-exported `AZURE_TENANT=$AZURE_TENANT_ID`.

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/module_utils/azure_rm_common.py
plugins/doc_fragments/azure_plugin.py

##### ADDITIONAL INFORMATION
1. `AZURE_TENANT_ID` honoured as a fallback for `tenant`
    - New `AZURE_CREDENTIAL_ENV_MAPPING_FALLBACK = {'tenant': 'AZURE_TENANT_ID'}`.
    - `_get_env()` returns the primary env var when set, otherwise the fallback.
    - `_get_env_credentials()` now routes through `_get_env()` so the alias applies there too.
    - `AZURE_TENANT` remains the primary name (back-compat with existing pipelines and `~/.azure/credentials` profiles).
2. `_get_env_credentials()` early-return reordered
    - Caller-supplied `subscription_id` is merged into `env_credentials` *before* the "no subscription return `None`" check, so a value from an `azure_rm.yml` inventory config or `~/.azure/credentials` profile is honoured under workload identity (where the webhook does not inject `AZURE_SUBSCRIPTION_ID`).
3. Auto cascade guard against env without auth material
    - `_get_env_credentials()` now returns `None` when the env has neither `client_id` (SP / OIDC paths) nor `ad_user` (user/password path), even if a caller-supplied `subscription_id` is present.
    - Without this guard, the change in (2) would cause `auth_source=auto` to short-circuit on a subscription-only env dict, skipping the documented `credential_file` -> `az-cli` fallbacks.
    - The guard preserves the #2274 fix because the AKS webhook always injects `AZURE_CLIENT_ID` alongside `AZURE_TENANT_ID` and `AZURE_FEDERATED_TOKEN_FILE`.

### DIFF
--- a/plugins/doc_fragments/azure.py
+++ b/plugins/doc_fragments/azure.py
@@ -159,7 +159,6 @@ notes:
     - Authentication is also possible using a service principal or Active Directory user.
     - To authenticate via OIDC, pass client_id, tenant and oidc_token or oidc_token_file_path, or set environment
       variables AZURE_CLIENT_ID, AZURE_TENANT and AZURE_FEDERATED_TOKEN or AZURE_FEDERATED_TOKEN_FILE.
-      AZURE_TENANT_ID is also accepted in place of AZURE_TENANT.
     - To authenticate via service principal, pass subscription_id, client_id, secret and tenant or set environment
       variables AZURE_SUBSCRIPTION_ID, AZURE_CLIENT_ID, AZURE_SECRET and AZURE_TENANT.
     - To authenticate via Active Directory user, pass ad_user and password, or set AZURE_AD_USER and
--- a/plugins/doc_fragments/azure.py
+++ b/plugins/doc_fragments/azure.py
@@ -159,6 +159,7 @@ notes:
     - Authentication is also possible using a service principal or Active Directory user.
     - To authenticate via OIDC, pass client_id, tenant and oidc_token or oidc_token_file_path, or set environment
       variables AZURE_CLIENT_ID, AZURE_TENANT and AZURE_FEDERATED_TOKEN or AZURE_FEDERATED_TOKEN_FILE.
+      AZURE_TENANT_ID is also accepted in place of AZURE_TENANT.
     - To authenticate via service principal, pass subscription_id, client_id, secret and tenant or set environment
       variables AZURE_SUBSCRIPTION_ID, AZURE_CLIENT_ID, AZURE_SECRET and AZURE_TENANT.
     - To authenticate via Active Directory user, pass ad_user and password, or set AZURE_AD_USER and
--- a/plugins/doc_fragments/azure_plugin.py
+++ b/plugins/doc_fragments/azure_plugin.py
@@ -98,7 +98,6 @@ notes:
     - Authentication is also possible using a service principal.
     - To authenticate via OIDC, pass client_id, tenant and oidc_token or oidc_token_file_path, or set environment
       variables AZURE_CLIENT_ID, AZURE_TENANT and AZURE_FEDERATED_TOKEN or AZURE_FEDERATED_TOKEN_FILE.
-      AZURE_TENANT_ID is also accepted in place of AZURE_TENANT.
     - To authenticate via service principal, pass subscription_id, client_id, secret and tenant or set environment
       variables AZURE_SUBSCRIPTION_ID, AZURE_CLIENT_ID, AZURE_SECRET and AZURE_TENANT.
     - "Alternatively, credentials can be stored in ~/.azure/credentials. This is an ini file containing
--- a/plugins/doc_fragments/azure_plugin.py
+++ b/plugins/doc_fragments/azure_plugin.py
@@ -36,9 +36,11 @@ options:
     tenant:
         description:
             - Azure tenant ID. Use when authenticating with a Service Principal.
+            - C(AZURE_TENANT_ID) is honoured as a fallback.
         type: str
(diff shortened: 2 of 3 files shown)


michielvha:
Hi @zunyangc thanks for the quick turnaround on this, really appreciate it !

i checked out the branch and ran the patched `_get_env` / `_get_env_credentials` against the exact env the AKS workload-identity webhook injects (`AZURE_CLIENT_ID` + `AZURE_TENANT_ID` + `AZURE_FEDERATED_TOKEN_FILE`, no `AZURE_TENANT` and no `AZURE_SUBSCRIPTION_ID`), passing the subscription the way an `azure_rm.yml` would. before the fix it returns `None` so auth fails, after the fix it resolves `tenant` from `AZURE_TENANT_ID` and merges the supplied `subscription_id`, so OIDC goes through.

the extra guard requiring `client_id` / `ad_user` before treating env as usable auth looks right too, the webhook always injects `AZURE_CLIENT_ID` so the WI path stays intact.

this is exactly what I needed downstream, I'd been bridging `AZURE_TENANT_ID` to `AZURE_TENANT` and can drop that once it's in a release. thanks again !

zunyangc:
Hi @Vieran, pipeline [tests](https://dev.azure.com/azclitools/internal/_build/results?buildId=321754&view=results) passed. Ready to merge, thanks!

Also completed smoke tests on different auth sources.

Vieran (approved):

PR #2278, opened 2026-06-12
Add databricks modules

p3ck:
##### SUMMARY
azure_rm_databricks_workspace* for creating/updating or deleting databrick workspaces.

Assisted-by: Claude

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_databricks_workspace.py
plugins/modules/azure_rm_databricks_workspace_info.py

##### ADDITIONAL INFORMATION
Integration tests pass

### DIFF
--- a/plugins/modules/azure_rm_databricks_workspace.py
+++ b/plugins/modules/azure_rm_databricks_workspace.py
@@ -135,6 +135,7 @@ EXAMPLES = '''
     resource_group: myResourceGroup
     name: myDatabricksWorkspace
     location: eastus
+    managed_resource_group_id: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx/resourceGroups/myManagedResourceGroup"
     sku:
       name: premium
     tags:
@@ -145,6 +146,7 @@ EXAMPLES = '''
     resource_group: myResourceGroup
     name: myDatabricksWorkspace
     location: eastus
+    managed_resource_group_id: "/subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxx/resourceGroups/myManagedResourceGroup"
     sku:
       name: premium
     parameters:
--- a/plugins/modules/azure_rm_databricks_workspace.py
+++ b/plugins/modules/azure_rm_databricks_workspace.py
@@ -52,14 +52,9 @@ options:
                     - standard
                     - premium
                     - trial
-            tier:
-                description:
-                    - The SKU tier.
-                type: str
     managed_resource_group_id:
         description:
             - The managed resource group ID.
-            - If not specified, a new resource group will be created automatically.
         type: str
     parameters:
         description:
@@ -199,7 +194,7 @@ sku:
         - The SKU of the workspace.
     type: dict
     returned: success
-    example: {'name': 'premium', 'tier': 'Premium'}
+    example: {'name': 'premium'}
 managed_resource_group_id:
     description:
         - The managed resource group ID.
@@ -252,7 +247,6 @@ class AzureRMDatabricksWorkspace(AzureRMModuleBaseExt):
                 type='dict',
                 options=dict(
                     name=dict(type='str', required=True, choices=['standard', 'premium', 'trial']),
-                    tier=dict(type='str')
                 )
             ),
             managed_resource_group_id=dict(type='str'),
@@ -316,7 +310,7 @@ class AzureRMDatabricksWorkspace(AzureRMModuleBaseExt):
             ('state', 'present', ['managed_resource_group_id']),
         ]
 
-        super(AzureRMDatabricksWorkspace, self).__init__(self.module_arg_spec, supports_check_mode=True, supports_tags=True)
+        super(AzureRMDatabricksWorkspace, self).__init__(self.module_arg_spec, supports_check_mode=True, supports_tags=True, required_if=required_if)
 
     def exec_module(self, **kwargs):
 
@@ -343,7 +337,6 @@ class AzureRMDatabricksWorkspace(AzureRMModuleBaseExt):
         elif workspace and self.state == 'present':
             update_tags, new_tags = self.update_tags(workspace.tags)
             if update_tags:
-                changed = True
                 self.tags = new_tags
 
             workspace_obj = self.create_workspace_obj()
@@ -363,7 +356,7 @@ class AzureRMDatabricksWorkspace(AzureRMModuleBaseExt):
                 self.delete_workspace()
 
         if workspace:
-            self.results = self.to_dict(workspace)
+            self.results = workspace.as_dict()
 
         self.results['changed'] = changed
         return self.results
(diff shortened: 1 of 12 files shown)


plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+                workspace_params['sku'] = self.databricks_models.Sku(
+                    name=self.sku.get('name'),
+                    tier=self.sku.get('tier')
+                )
+
+            if self.managed_resource_group_id:
+                workspace_params['managed_resource_group_id'] = self.managed_resource_group_id
+
+            if self.parameters:
+                workspace_params['parameters'] = {}
```
zunyangc: `parameters` is sent as a plain `dict`, every custom-parameter (i.e. VNet, NPIP, encryption, subnets) therefore get silently ignored.

The SDK's `Workspace._attribute_map['parameters']` is typed `WorkspaceCustomParameters`. `azure.mgmt.databricks 2.0.0` still uses the legacy `msrest Model.serialize()`, which only walks declared model types - a raw `dict` is dropped from the REST body.

plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+        self.location = None
+        self.sku = None
+        self.managed_resource_group_id = None
+        self.parameters = None
+
+        required_if = [
+            ('state', 'present', ['managed_resource_group_id']),
+        ]
+
+        super(AzureRMDatabricksWorkspace, self).__init__(self.module_arg_spec, supports_check_mode=True, supports_tags=True)
```
zunyangc:
```suggestion
        super(AzureRMDatabricksWorkspace, self).__init__(self.module_arg_spec, supports_check_mode=True, supports_tags=True, required_if=required_if)
```

plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+                    - premium
+                    - trial
+            tier:
+                description:
+                    - The SKU tier.
+                type: str
+    managed_resource_group_id:
+        description:
+            - The managed resource group ID.
+            - If not specified, a new resource group will be created automatically.
```
zunyangc: This seems incorrect: the SDK constructor requires `managed_resource_group_id`, and the `required_if` (once fixed in `__init__`) forbids omitting it.

plugins/modules/azure_rm_databricksworkspace_info.py
```diff
@@ -0,0 +1,222 @@
...
+        try:
+            return list(self.databricks_client.workspaces.list_by_subscription())
+        except Exception:
+            pass
+        return []
+
+    def to_dict(self, workspace):
+        result = workspace.as_dict()
+        result['resource_group'] = parse_resource_id(result['id'])['resource_group']
+        return result
```
zunyangc, line 214: Should we call `self.fail(...)` instead of silently swallowing `Exception`?

tests/integration/targets/azure_rm_databricks/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+      when: false  # Set to true if you have a VNet configured
+
+    # cleanup
+    - name: Delete the workspace
+      azure.azcollection.azure_rm_databricks_workspace:
+        name: "{{ workspace_name }}"
+        resource_group: "{{ resource_group }}"
+        state: absent
+      register: results
+
```
zunyangc, line 158: Missing idempotency check for delete.

plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+
+        if not workspace and self.state == 'present':
+            workspace_obj = self.create_workspace_obj()
+            changed = True
+            if not self.check_mode:
+                workspace = self.create_or_update_workspace(workspace_obj)
+        elif workspace and self.state == 'present':
+            update_tags, new_tags = self.update_tags(workspace.tags)
+            if update_tags:
+                changed = True
```
zunyangc: Is this line still needed as, it is immediately overwritten by https://github.com/ansible-collections/azure/pull/2278/changes#diff-9bb7ab797aaaa1ab72725766c94d0e44d52dfbd518715a82363ad87fd27ac729R351 a few lines later.

plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+        return None
+
+    def delete_workspace(self):
+        try:
+            poller = self.databricks_client.workspaces.begin_delete(self.resource_group, self.name)
+            return self.get_poller_result(poller)
+        except Exception as exc:
+            self.fail('Error deleting workspace {0} - {1}'.format(self.name, str(exc)))
+
+    def to_dict(self, workspace):
```
zunyangc: This is just a one-line wrapper, should we remove it and inline it at the call site?

plugins/modules/azure_rm_databricks_workspace.py
```diff
@@ -0,0 +1,459 @@
...
+                type: str
+                required: true
+                choices:
+                    - standard
+                    - premium
+                    - trial
+            tier:
+                description:
+                    - The SKU tier.
+                type: str
```
zunyangc: should we add `choices=['Standard', 'Premium', 'Trial']`?

tests/integration/targets/azure_rm_databricks/tasks/main.yml
```diff
@@ -0,0 +1,134 @@
...
+          custom_public_subnet_name:
+            value: public-subnet
+          custom_private_subnet_name:
+            value: private-subnet
+          enable_no_public_ip:
+            value: true
+        tags:
+          environment: production
+          purpose: analytics
+      when: false  # Set to true if you have a VNet configured
```
zunyangc: The `parameters: block` has no enabled test (this task is `when: false`). Please enable at least one parameters smoke test. The simplest one is `enable_no_public_ip` since it needs no pre-existing `VNet`.

zunyangc (changes requested):
Hi @p3ck, fyi thanks!

plugins/modules/azure_rm_databricks_workspace.py
p3ck: I looked closer at the examples in the command line tool, az databricks workspace create and they only specify the name.  name is the only required param so I am going to remove Tier since it just makes it more complicated where the Tier name must match the name.

plugins/modules/azure_rm_databricks_workspace_info.py
```diff
@@ -0,0 +1,222 @@
...
+            return list(self.databricks_client.workspaces.list_by_resource_group(self.resource_group))
+        except Exception as exc:
+            self.fail('Error listing workspaces by resource group {0} - {1}'.format(self.resource_group, str(exc)))
+        return []
+
+    def list_by_subscription(self):
+        try:
+            return list(self.databricks_client.workspaces.list_by_subscription())
+        except Exception as exc:
+            self.fail('Error listing workspaces by subscription - {1}'.format(str(exc)))
```
zunyangc:
```suggestion
            self.fail('Error listing workspaces by subscription - {0}'.format(str(exc)))
```

plugins/modules/azure_rm_databricks_workspace_info.py
```diff
@@ -0,0 +1,222 @@
...
+            response = self.list_by_subscription()
+
+        self.results['workspaces'] = [self.to_dict(x) for x in response if self.has_tags(x.tags, self.tags)]
+        return self.results
+
+    def get_workspace(self):
+        try:
+            return self.databricks_client.workspaces.get(self.resource_group, self.name)
+        except ResourceNotFoundError as exc:
+            self.fail('Error getting workspace {0} - {1}'.format(self.name, str(exc)))
```
zunyangc: Failing on "not found" will crash the module when creating a new workspace. Please revert

zunyangc (changes requested):
@p3ck, minor updates needed. Thanks!

plugins/modules/azure_rm_databricksworkspace.py
```diff
@@ -0,0 +1,448 @@
...
+      name: premium
+    parameters:
+      custom_virtual_network_id:
+        value: /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/myVnetRG/providers/Microsoft.Network/virtualNetworks/myVnet
+      custom_public_subnet_name:
+        value: public-subnet
+      custom_private_subnet_name:
+        value: private-subnet
+      enable_no_public_ip:
+        value: true
```
zunyangc, line 160: "Create" examples are missing `managed_resource_group_id` which is now required (`required_if`)

zunyangc (changes requested):

plugins/modules/azure_rm_databricks_workspace_info.py
p3ck: But we aren't creating a new workspace in _info module.  We are looking up an existing one.

zunyangc: The substantive point is convention — other `*_info` module in this collection treats `ResourceNotFoundError` as "no result", not as an error. 

Concrete scenario where this breaks today:

```yaml
- name: Check if Databricks workspace already exists
  azure_rm_databricks_workspace_info:
    resource_group: myRG
    name: myWS
  register: existing

- name: Create only if missing
  azure_rm_databricks_workspace:
    resource_group: myRG
    name: myWS
    location: eastus
    managed_resource_group_id: "/subscriptions/.../resourceGroups/myMRG"
    sku:
      name: premium
  when: existing.workspaces | length == 0
```

On a fresh resource group, the first task fails with `Error getting workspace myWS - (ResourceNotFound) ...` and the play never reaches the create. With `except ResourceNotFoundError: pass`, it would return `workspaces: []` and the conditional create works as users expect.

p3ck: I understand now. Thanks

zunyangc (dismissed):
LGTM thanks!

zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=323392&view=results) passed. Ready to merge, thanks!

Vieran (changes requested):
Should we add these 2 new modules to the `runtime.yml`?

zunyangc:
> Should we add these 2 new modules to the `runtime.yml`?

Added. The new modules are also renamed into:
`azure_rm_databricksworkspace`
`azure_rm_databricksworkspace_info`
to match the existing compound-name convention

@Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=326047&view=results) passed. Thanks!


_cc: @p3ck_

Vieran (approved):

PR #2287, opened 2026-06-24
Add new cognitive services account modules

p3ck:
##### SUMMARY
  - azure_rm_cognitiveservicesaccount - new module to manage Cognitive Services accounts
  - azure_rm_cognitiveservicesaccount_info - new module to query Cognitive Servies account i nformation

##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
azure_rm_cognitiveservicesaccount
azure_rm_cognitiveservicesaccount_info

##### ADDITIONAL INFORMATION

None of the examples in the python azure SDK use the model when building the request for Azure.  And as you can see the integration tests test that we are creating the service correctly.

Thanks

### DIFF
--- a/plugins/modules/azure_rm_cognitiveservicesaccount.py
+++ b/plugins/modules/azure_rm_cognitiveservicesaccount.py
@@ -25,7 +25,7 @@ options:
     name:
         description:
             - Name of the Cognitive Services account.
-            - Must be 2-64 characters, alphanumeric, hyphens, underscores.
+            - Must be 2-64 characters, alphanumeric, hyphens, underscores and dots.
             - Must be globally unique.
         required: true
         type: str
@@ -76,7 +76,7 @@ options:
             - The SKU/pricing tier of the Cognitive Services account.
             - C(F0) is the free tier with rate limits.
             - C(S0) is the standard tier.
-        default: F0
+            - If not specified on Create F0 is the default.
         type: str
         choices:
             - F0
@@ -88,7 +88,7 @@ options:
     public_network_access:
         description:
             - Whether public network access is allowed.
-        default: Enabled
+            - Defaults to Enabled on Creation.
         type: str
         choices:
             - Enabled
@@ -96,7 +96,7 @@ options:
     disable_local_auth:
         description:
             - Disable key-based authentication, require Microsoft Entra ID.
-        default: false
+            - Defaults to False on Creation.
         type: bool
     network_acls:
         description:
@@ -147,6 +147,11 @@ options:
                     - SystemAssigned
                     - UserAssigned
                     - SystemAssigned, UserAssigned
+            user_assigned_identities:
+                description:
+                    - User Assigned Identities.
+                type: list
+                elements: str
     purge:
         description:
             - When I(state) is C(absent) setting this to true
@@ -373,14 +378,11 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
                        choices=['present', 'absent']),
             location=dict(type='str'),
             sku=dict(type='str',
-                     default='F0',
                      choices=['F0', 'S0']),
             custom_domain_name=dict(type='str'),
             public_network_access=dict(type='str',
-                                       default='Enabled',
                                        choices=['Enabled', 'Disabled']),
-            disable_local_auth=dict(type='bool',
-                                    default=False),
+            disable_local_auth=dict(type='bool'),
             purge=dict(type='bool',
                        default=False),
             network_acls=dict(
@@ -413,7 +415,11 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
                                        'SystemAssigned',
                                        'UserAssigned',
                                        'SystemAssigned, UserAssigned']
-                              )
+                              ),
+                    user_assigned_identities=dict(type='list',
+                                                  elements='str',
+                                                  required=False,
+                                                  ),
                 )
             ),
             tags=dict(type='dict')
@@ -459,12 +465,18 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         # Get existing account
         account = self.get_account()
 
-        params = self.build_account_parameters()
-
         if self.state == 'present':
             if not account:
+                if not self.sku:
+                    self.sku = 'F0'
+                if not self.public_network_access:
+                    self.public_network_access = 'Enabled'
+                if self.disable_local_auth is None:
+                    self.disable_local_auth = False
                 if not self.kind:
                     self.module.fail_json(msg="kind must be specified when creating")
+                params = self.build_account_parameters()
+
                 # Create new account
                 if not self.check_mode:
                     self.results['state'] = self.create_account(params)
@@ -473,6 +485,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
                 self.results['changed'] = True
             else:
                 # Update existing account
+                params = self.build_account_parameters()
                 update_needed = self.check_update_needed(account, params)
                 if update_needed:
                     if not self.check_mode:
@@ -485,7 +498,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         else:  # state == 'absent'
             if account:
                 if not self.check_mode:
-                    self.delete_account()
+                    self.delete_account(account['location'])
                 self.results['changed'] = True
                 self.results['state'] = dict()
             else:
@@ -523,12 +536,6 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         update_tags, dummy = self.update_tags(account.get('tags'))
         if update_tags:
             return True
-
-        # Check identity
-        if self.identity:
-            account_identity_type = account.get('identity', {}).get('type', 'None')
-            if account_identity_type != self.identity.get('type', 'None'):
-                return True
 
         if not self.default_compare({},
                                     params,
@@ -587,6 +594,12 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
                 'type': self.identity['type']
             }
 
+            userAssignedTypes = ['UserAssigned', 'SystemAssigned, UserAssigned']
+            if self.identity.get('type') in userAssignedTypes and \
+                    self.identity.get('user_assigned_identities'):
+                identities = {identity: dict() for identity in self.identity['user_assigned_identities']}
+                params['identity']['user_assigned_identities'] = identities
+
         # Tags
         params['tags'] = self.tags
 
@@ -615,10 +628,10 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         params['kind'] = current_account['kind']
 
         # Update SKU if specified and different
-        if self.sku and current_account['sku'] != self.sku:
+        if self.sku and current_account['sku']['name'] != self.sku:
             params['sku'] = {'name': self.sku}
         else:
-            params['sku'] = {'name': current_account['sku']}
+            params['sku'] = {'name': current_account['sku']['name']}
 
         # Handle tags
         update_tags, new_tags = self.update_tags(current_account.get('tags'))
@@ -635,7 +648,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         except Exception as exc:
             self.module.fail_json(msg="Failed to update Cognitive Services account: {0}".format(str(exc)))
 
-    def delete_account(self):
+    def delete_account(self, location):
         """Delete Cognitive Services account"""
         self.log('Deleting Cognitive Services account {0}'.format(self.name))
 
@@ -650,7 +663,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         if self.purge:
             try:
                 poller = self.cognitive_services_management_client.deleted_accounts.begin_purge(
-                    self.location,
+                    location,
                     self.resource_group,
                     self.name
                 )
--- a/plugins/modules/azure_rm_cognitiveservicesaccount.py
+++ b/plugins/modules/azure_rm_cognitiveservicesaccount.py
@@ -10,7 +10,7 @@ __metaclass__ = type
 DOCUMENTATION = '''
 ---
 module: azure_rm_cognitiveservicesaccount
-version_added: "3.20.0"
+version_added: "3.21.0"
 short_description: Manage Azure AI Cognitive Services accounts
 description:
     - Create, update, or delete Azure AI Cognitive Services accounts.
@@ -133,25 +133,6 @@ options:
                             - Full resource ID of the subnet.
                         type: str
                         required: true
-    identity:
-        description:
-            - Managed identity configuration.
-        type: dict
-        suboptions:
-            type:
-                description:
-                    - Type of managed identity.
-                type: str
-                choices:
-                    - None
-                    - SystemAssigned
-                    - UserAssigned
-                    - SystemAssigned, UserAssigned
-            user_assigned_identities:
-                description:
-                    - User Assigned Identities.
-                type: list
-                elements: str
     purge:
         description:
             - When I(state) is C(absent) setting this to true
@@ -162,6 +143,7 @@ options:
 extends_documentation_fragment:
     - azure.azcollection.azure
     - azure.azcollection.azure_tags
+    - azure.azcollection.azure_identity_multiple_user
 
 author:
     - Bill Peck (@p3ck)
@@ -334,6 +316,7 @@ import re
 from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common_ext import AzureRMModuleBaseExt
 
 try:
+    from azure.mgmt.cognitiveservices.models import Identity
     from azure.core.exceptions import ResourceNotFoundError
 except ImportError:
     # This is handled in azure_rm_common
@@ -409,18 +392,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
             ),
             identity=dict(
                 type='dict',
-                options=dict(
-                    type=dict(type='str',
-                              choices=['None',
-                                       'SystemAssigned',
-                                       'UserAssigned',
-                                       'SystemAssigned, UserAssigned']
-                              ),
-                    user_assigned_identities=dict(type='list',
-                                                  elements='str',
-                                                  required=False,
-                                                  ),
-                )
+                options=self.managed_identity_multiple_user_assigned_spec
             ),
             tags=dict(type='dict')
         )
@@ -437,6 +409,7 @@ class AzureRMCognitiveServicesAccount(AzureRMModuleBaseExt):
         self.identity = None
         self.tags = None
         self.purge = None
+        self._managed_identity = None
 
         self.results = dict(changed=False,
                             compare=[])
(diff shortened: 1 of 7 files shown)


plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+            - F0
+            - S0
+    custom_domain_name:
+        description:
+            - Optional custom subdomain name for token-based authentication.
+        type: str
+    public_network_access:
+        description:
+            - Whether public network access is allowed.
+        default: Enabled
```
zunyangc: User who created the account with `public_network_access: Disabled` and re-runs the task without re-specifying it will silently revert it to `Enabled`. Same for [`disable_local_auth`](https://github.com/ansible-collections/azure/pull/2287/changes#diff-ae45ec7c97fdb2bfb960f237b3e2c9e52e0b8f4e39de8779883a4a10b8b5f452R99).

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+        type: str
+        choices:
+            - Enabled
+            - Disabled
+    disable_local_auth:
+        description:
+            - Disable key-based authentication, require Microsoft Entra ID.
+        default: false
```
zunyangc: Same as `public_network_access`.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+        self.log('Updating Cognitive Services account {0}'.format(self.name))
+
+        # Preserve immutable fields from current account
+        params['kind'] = current_account['kind']
+
+        # Update SKU if specified and different
+        if self.sku and current_account['sku'] != self.sku:
+            params['sku'] = {'name': self.sku}
+        else:
+            params['sku'] = {'name': current_account['sku']}
```
zunyangc: `current_account['sku']` is a `dict` but `self.sku` is a string, `else` branch is dead. Should it be `current_account['sku']['name'] != self.sku`, with the else branch using `current_account['sku']['name']` ?

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+        description:
+            - Valid Azure location.
+            - Defaults to location of the resource group.
+        type: str
+    sku:
+        description:
+            - The SKU/pricing tier of the Cognitive Services account.
+            - C(F0) is the free tier with rate limits.
+            - C(S0) is the standard tier.
+        default: F0
```
zunyangc: Because `sku` has `default`, every invocation sends `F0`, causing update when re-runs without re-specifying the SKU.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+        type: dict
+        suboptions:
+            type:
+                description:
+                    - Type of managed identity.
+                type: str
+                choices:
+                    - None
+                    - SystemAssigned
+                    - UserAssigned
```
zunyangc: Arg-spec has no `user_assigned_identities` sub-option?

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+                self.module.fail_json(msg="Cannot change SKU from {0} to {1}".format(account['sku']['name'], self.sku))
+            return True
+
+        # Check tags
+        update_tags, dummy = self.update_tags(account.get('tags'))
+        if update_tags:
+            return True
+
+        # Check identity
+        if self.identity:
```
zunyangc: `check_update_needed` only returns True when `self.identity` is set and its `type differs`.
2 cases miss:
- user removed `identity`
- (the user-assigned IDs set also needs diffing. )

plugins/modules/azure_rm_cognitiveservicesaccount_info.py
```diff
@@ -0,0 +1,319 @@
...
+from ansible_collections.azure.azcollection.plugins.module_utils.azure_rm_common import AzureRMModuleBase
+
+try:
+    from azure.core.exceptions import ResourceNotFoundError
+except ImportError:
+    # This is handled in azure_rm_common
+    pass
+
+
+class AzureRMAICognitiveServicesAccountInfo(AzureRMModuleBase):
```
zunyangc:
```suggestion
class AzureRMCognitiveServicesAccountInfo(AzureRMModuleBase):
```

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+                self.resource_group,
+                self.name
+            )
+            self.get_poller_result(poller)
+        except Exception as exc:
+            self.module.fail_json(msg="Failed to delete Cognitive Services account: {0}".format(str(exc)))
+        if self.purge:
+            try:
+                poller = self.cognitive_services_management_client.deleted_accounts.begin_purge(
+                    self.location,
```
zunyangc: `begin_purge` called with `self.location`, which falls back to the rg's default location when the user doesn't pass `location` on the delete task. If the account was created in a different region than the RG, this silently 404s. Please capture the location from the fetched `account` object before calling `delete_account()` and pass that into `begin_purge`.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,667 @@
...
+        description:
+            - Name of the resource group.
+        required: true
+        type: str
+        aliases:
+            - resource_group_name
+    name:
+        description:
+            - Name of the Cognitive Services account.
+            - Must be 2-64 characters, alphanumeric, hyphens, underscores.
```
zunyangc: `dots` is also allowed. Should we add dots to the description for accuracy.

plugins/modules/azure_rm_cognitiveservicesaccount_info.py
```diff
@@ -0,0 +1,319 @@
...
+                resource_group,
+                account_name
+            )
+            return {
+                'key1': keys_obj.key1,
+                'key2': keys_obj.key2
+            }
+        except Exception as exc:
+            self.log('Error retrieving keys: {0}'.format(str(exc)))
+            return {}
```
zunyangc: `get_keys` swallow exception and returns {}. A caller with `Reader` but without `Microsoft.CognitiveServices/accounts/listKeys/action` will get a 403 here. With `show_keys: true` they'll see `auth_keys: {}` in the output with no diagnostic. Should we call `self.fail(...)` on error?

zunyangc (changes requested):
Hi @p3ck, FYI. 

Should we also add new modules into action group entries? In `meta/runtime.yml` (Just found out about this).

Thanks!

plugins/modules/azure_rm_cognitiveservicesaccount.py
p3ck: The user_assigned identities will show up in the default_compare

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,690 @@
...
+
+        # Update SKU if specified and different
+        if self.sku and current_account['sku']['name'] != self.sku:
+            params['sku'] = {'name': self.sku}
+        else:
+            params['sku'] = {'name': current_account['sku']['name']}
+
+        # Handle identity
+        if not self.identity and 'identity' in current_account:
+            params['identity'] = current_account['identity']
```
zunyangc: When the user omits `identity` but the existing account has one, this will copy `current_account['identity']` verbatim into the update payload. That dict includes read-only fields (`principal_id`, `tenant_id`, and etc. ) which shouldn't be echoed back to `begin_update`.

plugins/modules/azure_rm_cognitiveservicesaccount.py
zunyangc: But `default_compare` in `azure_rm_common_ext` treats a key present in `old` but missing in `new` as "incremental_update" and silently copies the old value back into `new` without flipping the result flag. 

So if the user drops UAI from the list, the module reports `changed=false` and never sends the update to the collection's shared `managed_identity_multiple_spec` + `self.update_managed_identity(..., patch_support=True)` helper in `plugins/module_utils/azure_rm_common_ext.py`.

tests/integration/targets/azure_rm_cognitiveservicesaccount/tasks/main.yml
```diff
@@ -0,0 +1,411 @@
...
+
+    - name: Assert no change on second delete
+      ansible.builtin.assert:
+        that:
+          - delete_again.changed == false
+
+    - name: Delete user managed identities
+      ansible.builtin.include_tasks: "{{ role_path }}/../../../integration_common_tasks/managed_identity.yml"
+      vars:
+        managed_identity_test_unique: 'csa'
```
zunyangc: Should we move this into the `always:` block?

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,690 @@
...
+#
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = '''
+---
+module: azure_rm_cognitiveservicesaccount
+version_added: "3.20.0"
```
zunyangc:
```suggestion
version_added: "3.21.0"
```

zunyangc (changes requested):

plugins/modules/azure_rm_cognitiveservicesaccount.py
p3ck: Good catch.  I realize this isn't needed at all.  I removed it and added a test case to verify.

plugins/module_utils/azure_rm_common_ext.py
```diff
@@ -340,7 +346,7 @@ def update_managed_identity(self, new_identity, curr_identity=None,
         # If type in module args contains 'UserAssigned'
```
zunyangc, line 340: Regression:
After reassigning `new_user_identities=set(...)`, calling `.get('append', True)` on line 335 raises `AttributeError: 'set' object has no attribute 'get'`. This is the hot path for every existing module using the shared `managed_identity_multiple_spec`. Please fix by reading `append` before the reassignment.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,686 @@
...
+            if self.disable_local_auth is not None:
+                properties['disable_local_auth'] = self.disable_local_auth
+        else:
+            if not self.kind:
+                self.module.fail_json(msg="kind must be specified when creating")
+            params['kind'] = self.kind
+            update_tags, params['tags'] = self.update_tags(None)
+            params['sku'] = {'name': self.sku or 'F0'}
+            properties['public_network_access'] = self.public_network_access or 'Enabled'
+            properties['disable_local_auth'] = self.disable_local_auth or False
```
zunyangc:
```suggestion
            properties['disable_local_auth'] = self.disable_local_auth if self.disable_local_auth is not None else False
```

tests/integration/targets/azure_rm_cognitiveservicesaccount/tasks/main.yml
```diff
@@ -0,0 +1,428 @@
...
+      azure.azcollection.azure_rm_cognitiveservicesaccount:
+        resource_group: "{{ resource_group }}-csa"
+        name: "{{ account_name }}"
+        kind: ContentSafety
+        sku: F0
+        location: eastus
+        public_network_access: Disabled
+      register: identity_result
+
+    - name: Assert identity did not Change
```
zunyangc:
```suggestion
    - name: Assert account changed but identity preserved
```

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,686 @@
...
+                self.module.fail_json(msg="Cannot change SKU from {0} to {1}".format(account['sku']['name'], params['sku']['name']))
+            changed = True
+
+        # Check tags
+        update_tags, params['tags'] = self.update_tags(account.get('tags'))
+        if update_tags:
+            changed = True
+
+        # Remove identity since we will check for that special case below.
+        old_identity = account.pop('identity', None)
```
zunyangc: This will mutate the dict returned by `get_account()`. When `check_update_needed` returns False, `exec_module` returns `self.results['state'] = account` with `identity` missing. User doing `result.state.identity.type` will hit KeyError on idempotent reruns but success on changed reruns.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,686 @@
...
+
+        # Check identity
+        if self.identity:
+            update_identity, identity = self.update_managed_identity(
+                self.identity,
+                old_identity,
+                patch_support=True
+            )
+            params['identity'] = dict(type=identity.type)
+            params['identity']['user_assigned_identities'] = identity.user_assigned_identities
```
zunyangc: Consistency nit:
When `type=SystemAssigned, the `identity.user_assigned_identities` now become None, so the payload sent to `begin_update` contains `userAssignedIdentities: null`. While Azure appears to accept null today, this could be a consideration:
```python
params['identity'] = dict(type=identity.type)
if identity.user_assigned_identities:
    params['identity']['user_assigned_identities'] = identity.user_assigned_identities
```

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,686 @@
...
+                    - List of allowed virtual network subnets.
+                type: list
+                elements: dict
+                suboptions:
+                    id:
+                        description:
+                            - Full resource ID of the subnet.
+                        type: str
+                        required: true
+    identity:
```
zunyangc: Should we align the `identity` arg-spec by using `options=self.managed_identity_multiple_spec` from the `azure_rm_common_ext`?

zunyangc (changes requested):

p3ck:
I updated to use the standard multiple user identities from module_utils.

zunyangc (dismissed):
LGTM, thanks!

zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=328259&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,656 @@
...
+        # Validate account name
+        if not re.match(r'^[a-zA-Z0-9][a-zA-Z0-9_.-]*$', self.name):
+            self.module.fail_json(msg="Account name must start with alphanumeric and contain only alphanumeric, hyphens, underscores, and dots")
+        if len(self.name) < 2 or len(self.name) > 64:
+            self.module.fail_json(msg="Account name must be between 2 and 64 characters")
+
+        # Get resource group location if location not specified
+        resource_group = self.get_resource_group(self.resource_group)
+        if not self.location:
+            self.location = resource_group.location
```
Vieran: Resource group location is not necessarily the resource location. Consider the scenario that the user want to update the resource but not pass the location, this should default to the current account location, or only enforce the immutable-location check when the user actually supplied location.

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,656 @@
...
+
+            if 'ip_rules' in self.network_acls and self.network_acls['ip_rules']:
+                network_rule_set['ip_rules'] = [
+                    {'value': rule['value']} for rule in self.network_acls['ip_rules']
+                ]
+
+            if 'virtual_network_rules' in self.network_acls and self.network_acls['virtual_network_rules']:
+                network_rule_set['virtual_network_rules'] = [
+                    {'id': rule['id']} for rule in self.network_acls['virtual_network_rules']
+                ]
```
Vieran: `ip_rules` and `virtual_network_rules` are only added when the list is truthy, so an explicit empty list is dropped from the update payload. That means users cannot clear existing ACL rules with `ip_rules: []` or `virtual_network_rules: []`; the task may look idempotent while old rules remain.

Could we check `is not None` instead, so omitted fields stay untouched but explicit empty lists are sent to Azure?

plugins/modules/azure_rm_cognitiveservicesaccount_info.py
```diff
@@ -0,0 +1,318 @@
...
+                if self.show_keys:
+                    # Extract resource group from account ID
+                    # Format: /subscriptions/{sub}/resourceGroups/{rg}/providers/...
+                    id_parts = account_obj.id.split('/')
+                    rg_index = id_parts.index('resourceGroups') + 1
+                    account_dict['auth_keys'] = self.get_keys(id_parts[rg_index],
+                                                              account_obj.name)
+                results.append(account_dict)
+        except Exception as exc:
+            self.log('Error listing accounts: {0}'.format(str(exc)))
```
Vieran: Same as above.

plugins/modules/azure_rm_cognitiveservicesaccount_info.py
```diff
@@ -0,0 +1,318 @@
...
+                self.resource_group
+            )
+            for account_obj in accounts:
+                account_dict = account_obj.as_dict()
+                if self.show_keys:
+                    account_dict['auth_keys'] = self.get_keys(self.resource_group,
+                                                              account_obj.name)
+                results.append(account_dict)
+        except Exception as exc:
+            self.log('Error listing accounts: {0}'.format(str(exc)))
```
Vieran: Please don't swallow list failures here. If the SDK call fails due to auth, permissions, provider errors, or networking, this returns `accounts: []`, which is indistinguishable from a genuinely empty result.

These list paths should fail the module with the original error; `[]` should only be used for the specific-account `ResourceNotFoundError` case.

Vieran (changes requested):

plugins/modules/azure_rm_cognitiveservicesaccount.py
```diff
@@ -0,0 +1,665 @@
...
+                properties['public_network_access'] = self.public_network_access
+            if self.disable_local_auth is not None:
+                properties['disable_local_auth'] = self.disable_local_auth
+        else:
+            if self.location:
+                params['location'] = self.location
+            else:
+                # Get resource group location if location not specified
+                resource_group = self.get_resource_group(self.resource_group)
+                self.location = resource_group.location
```
Vieran: If the branch goes to "else", the `params['location']` will not be set in this code.

Should we write something like
```
if not self.location:
    resource_group = self.get_resource_group(self.resource_group)
    self.location = resource_group.location
params['location'] = self.location
```

Vieran (changes requested):

plugins/modules/azure_rm_cognitiveservicesaccount.py
p3ck: Yes - Thank you for catching this.  I don't know why I'm having such a hard time lately.

Vieran (approved):

PR #2288, opened 2026-06-25
Fix `azure_rm_virtualmachine`: default to managed disks (unmanaged disks retired 2026-03-31)

zunyangc:
##### SUMMARY

Azure retired unmanaged disks (VHD page blobs attached to VMs) on **March 31, 2026**. The VM control plane now rejects any create/update that references a VHD blob for the OS or data disks with:
```
(BadRequest) Unmanaged disks have been retired. Please migrate to Managed Disks.
```
`azure_rm_virtualmachine` previously defaulted to the unmanaged code path whenever the caller specified neither `managed_disk_type` nor `storage_blob_name`/ `storage_account_name`. This broke every "minimal VM" task across this collection's integration tests (and every downstream user's playbook that doesn't explicitly opt into managed disks).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_virtualmachine.py

##### ADDITIONAL INFORMATION
- **OS disk** — when no disk options are provided, default to managed `Standard_LRS` instead of falling back to a VHD blob.
- **Data disks** — same fallback inside the `data_disks` loop, per `data_disk` entry.
- **Deprecation warning** — when the caller explicitly opts into the unmanaged path (passes `storage_blob_name`, `storage_account_name`, or `storage_container_name` at either the OS-disk or `data_disks` level), emit a runtime `module.deprecate(..., version='4.0.0', collection_name='azure.azcollection')` pointing at the Microsoft retirement notice. 
- **DOCUMENTATION** — tagged the six affected parameters with `(Deprecated)` and added the retirement-notice URL as a bullet.

----
**Reference**
- https://learn.microsoft.com/en-us/azure/virtual-machines/unmanaged-disks-deprecation

### DIFF
--- a/plugins/modules/azure_rm_virtualmachine.py
+++ b/plugins/modules/azure_rm_virtualmachine.py
@@ -199,24 +199,33 @@ options:
                 type: str
     storage_account_name:
         description:
-            - Name of a storage account that supports creation of VHD blobs.
+            - (Deprecated) Name of a storage account that supports creation of VHD blobs.
             - If not specified for a new VM, a new storage account named <vm name>01 will be created using storage type C(Standard_LRS).
+            - Unmanaged disks were retired by Azure on March 31, 2026. Use I(managed_disk_type) instead.
+            - See U(https://learn.microsoft.com/azure/virtual-machines/unmanaged-disks-deprecation).
+            - This option will be removed in version 4.0.0.
         type: str
         aliases:
             - storage_account
     storage_container_name:
         description:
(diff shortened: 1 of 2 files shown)


zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=325684&view=results) passed. Ready to merge, thanks!

Vieran (approved):

PR #2299, opened 2026-07-15
fix: add UUID validation to Azure EDA source plugins

ttuffin:
Both plugins were writing to meta["event"]["uuid"] — a nested sub-dict that ansible-rulebook's insert_meta_info filter never reads. This caused every event to receive a random UUID4 regardless of what the plugin set, and eda-server validation failures when the Azure-supplied ID was not a valid RFC 4122 UUID.

Fixes:
- Set meta["uuid"] (flat) to match the schema expected by ansible-rulebook and eda-server
- Validate Azure-supplied IDs before use; fall back to a deterministic UUID5 when invalid, preserving the original under meta["message_id"]
- Move produced_at from meta["event"] to meta (flat) for consistency with the Kafka plugin
- Add one-time warning when an invalid ID triggers fallback

Added test cases and refactored existing tests for de-duplication.

UUID5 seeds:
- azure_service_bus: queue_name:sequence_number
- azure_event_hub: partition_id:sequence_number coordinates

Refs: AAP-77077, AAP-77078

Assisted-by: Claude

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
EDA

### DIFF
--- a/extensions/eda/plugins/event_source/azure_event_hub.py
+++ b/extensions/eda/plugins/event_source/azure_event_hub.py
@@ -7,6 +7,7 @@ events into Ansible EDA rulebooks.
 import asyncio
 import json
 import logging
+import uuid
 from typing import Any
 
 from azure.eventhub import EventData
@@ -19,6 +20,14 @@ short_description: Receive events via a Azure Event Hub
 description:
   - An ansible-rulebook event source plugin for receiving events
     via Azure Event Hub
+  - Each event is assigned a valid RFC 4122 UUID under meta.uuid.
+  - If message_id is a valid UUID it is used. Otherwise if correlation_id is a valid UUID
+    it is used. Each is validated independently so an invalid message_id does not hide
+    a valid correlation_id.
+  - AMQP ID values of any type (str, bytes, uuid.UUID, int) are normalised to str
+    before validation.
+  - If neither is a valid UUID, a deterministic UUID5 is generated from
+    partition_id:sequence_number. The original value is preserved under meta.message_id.
 options:
   azure_tenant_id:
     description:
@@ -94,6 +103,74 @@ EXAMPLES = r"""
 
 logger = logging.getLogger()
 
+_uuid_warning = {"emitted": False}
+
+
+def is_valid_uuid(value: str) -> bool:
+    """Check if a string is a valid RFC 4122 UUID.
+
+    Accepts a str but handles non-string types defensively by catching
+    TypeError, so callers do not need to guard against unexpected types.
+    """
+    try:
+        uuid.UUID(value)
+        return True  # noqa: TRY300
+    except (ValueError, AttributeError, TypeError):
+        return False
+
+
+def _normalize_amqp_id(value: str | bytes | uuid.UUID | int | None) -> str | None:
+    """Normalize an AMQP ID value to a string suitable for UUID validation.
+
+    AMQP message-id and correlation-id may be str, bytes, uuid.UUID, or int (ulong).
+    Returns None if value is None.
+    """
+    if value is None:
+        return None
+    if isinstance(value, bytes):
+        try:
+            return value.decode("utf-8")
+        except UnicodeDecodeError:
+            return value.hex()
+    return str(value)
+
+
+def _process_event_hub_uuid(
+    partition_context: PartitionContext,
+    event: EventData,
+    meta: dict[str, Any],
+) -> None:
+    """Assign a valid UUID to meta["uuid"].
+
+    Priority: message_id > correlation_id > generated UUID5.
+    Each candidate is validated independently so an invalid message_id does
+    not hide a valid correlation_id. Generated UUID5 uses
+    partition_id:sequence_number as coordinates. The original value is
+    preserved under meta["message_id"] when the fallback is used.
+    """
+    coordinates = f"{partition_context.partition_id}:{event.sequence_number}"
+    generated_uuid = str(uuid.uuid5(uuid.NAMESPACE_OID, coordinates))
+
+    message_id = _normalize_amqp_id(event.message_id)
+    correlation_id = _normalize_amqp_id(event.correlation_id)
+
+    if message_id and is_valid_uuid(message_id):
+        meta["uuid"] = message_id
+    elif correlation_id and is_valid_uuid(correlation_id):
+        meta["uuid"] = correlation_id
+    else:
+        meta["uuid"] = generated_uuid
+        raw_id = message_id or correlation_id
+        if raw_id:
+            meta["message_id"] = raw_id
+            if not _uuid_warning["emitted"]:
+                _uuid_warning["emitted"] = True
+                logger.warning(
+                    "Provided message_id/correlation_id is not a valid UUID,"
+                    " using generated UUID. The original value has been"
+                    " stored under event.meta.message_id for tracking.",
+                )
+
 
 REQUIRED_ARGS = [
     "azure_tenant_id",
@@ -144,16 +221,10 @@ class AzureHubConsumer:
     ) -> None:
         """Receiving event data."""
         if event:
-            meta = {}
-            msg_id = (
-                event.message_id
-                or event.correlation_id
-                or f"{partition_context.partition_id}:{event.sequence_number}"
-            )
-
-            meta["event"] = {"uuid": msg_id}
+            meta: dict[str, Any] = {}
             if event.enqueued_time:
-                meta["event"]["produced_at"] = event.enqueued_time.isoformat()
+                meta["produced_at"] = event.enqueued_time.isoformat()
+            _process_event_hub_uuid(partition_context, event, meta)
             # Process message body
             try:
                 value = event.body_as_str()
--- a/extensions/eda/plugins/event_source/azure_service_bus.py
+++ b/extensions/eda/plugins/event_source/azure_service_bus.py
@@ -8,8 +8,10 @@ import asyncio
 import contextlib
 import json
 import logging
+import uuid
 from typing import Any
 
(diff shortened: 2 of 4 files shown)


mkanoor (approved):

mkanoor:
@p3ck Can you please review

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -71,6 +77,42 @@
...
+
+
+def is_valid_uuid(value: str) -> bool:
+    """Check if a string is a valid RFC 4122 UUID."""
+    try:
+        uuid.UUID(value)
+        return True  # noqa: TRY300
+    except (ValueError, AttributeError):
+        return False
+
```
p3ck, line 95: I'm not sure how to share code from the extensions directory but if it's possible I think we should move these two methods there.

AlexSCorey (approved):
looks good to me.

extensions/eda/plugins/event_source/azure_service_bus.py
ttuffin: Let me run this by @mkanoor. As far as I know there is no precedent for sharing code between source plugins.

mkanoor: @p3ck Currently when we run the source plugins via ansible-rulebook, there is no explicit setting of the python system path. We use the runpy.run_path which requires that everything be self contained or use system installed python packages. Since collections get installed in different locations from standard python packages we need to tweak the paths before we call the plugins. This is mostly needed for code sharing between modules and more manageable source plugin code. We are looking into this with Dafne and have plans to implement the module_utils directory with shared code akin to ansible collections. At that point we can move these into a shared module.

p3ck (dismissed):
Thanks!

zunyangc (dismissed):
LGTM, thanks!

zunyangc:
Hi @Vieran, pipeline [tests](https://dev.azure.com/azclitools/internal/_build/results?buildId=330435&view=results) passed. Ready to merge, thanks!

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -71,6 +77,42 @@
...
+def _process_service_bus_uuid(msg: ServiceBusReceivedMessage, meta: dict[str, Any]) -> None:
+    """Assign a valid UUID to meta["uuid"].
+
+    Uses msg.message_id if it is a valid UUID, otherwise generates a
+    deterministic UUID5 from the message_id string. The original value
+    is preserved under meta["message_id"] when it is not a valid UUID.
+    """
+    raw_id = str(msg.message_id) if msg.message_id is not None else None
+    generated_uuid = str(uuid.uuid5(uuid.NAMESPACE_OID, raw_id or ""))
```
Vieran: The UUID5 seed contains only `message_id`, or an empty string when it is missing. Therefore, messages with the same ID from different queues/namespaces receive the same UUID, and all messages without an ID receive one identical UUID. Since `meta.uuid` is used as the downstream event identity, this can incorrectly merge unrelated events.

Please include the source identity and a broker-unique value such as `sequence_number` in the seed, without including secrets from the connection string.

extensions/eda/plugins/event_source/azure_event_hub.py
```diff
@@ -94,6 +99,49 @@
...
+_uuid_warning = {"emitted": False}
+
+
+def is_valid_uuid(value: str) -> bool:
+    """Check if a string is a valid RFC 4122 UUID."""
+    try:
+        uuid.UUID(value)
+        return True  # noqa: TRY300
+    except (ValueError, AttributeError):
+        return False
```
Vieran, line 119: AMQP `message-id` and `correlation-id` values may be `ulong`, `UUID`, `binary`, or `string`, and the Azure SDK can preserve these incoming types. `uuid.UUID(value)` assumes a string here: a binary value can raise an uncaught `TypeError`, stopping consumption before the event is queued or checkpointed.

Please normalize the ID to a canonical, JSON-safe string before validation and add coverage for `bytes`, `uuid.UUID`, and integer IDs.

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -71,6 +77,42 @@
...
+
+    if raw_id and is_valid_uuid(raw_id):
+        meta["uuid"] = raw_id
+    else:
+        meta["uuid"] = generated_uuid
+        if raw_id:
+            meta["message_id"] = raw_id
```
Vieran: Previously, every Service Bus event included `meta.message_id`, but this code now sets it only when the ID is invalid. Since the Azure SDK normally generates a valid UUID, most events will lose this existing field and rules using `event.meta.message_id` may break.

Please preserve the original `message_id` in meta regardless of whether it is already a valid UUID.

extensions/eda/plugins/event_source/azure_event_hub.py
```diff
@@ -94,6 +99,49 @@
...
+    """
+    coordinates = f"{partition_context.partition_id}:{event.sequence_number}"
+    generated_uuid = str(uuid.uuid5(uuid.NAMESPACE_OID, coordinates))
+
+    raw_id = event.message_id or event.correlation_id
+
+    if raw_id and is_valid_uuid(raw_id):
+        meta["uuid"] = raw_id
+    else:
+        meta["uuid"] = generated_uuid
```
Vieran: `event.message_id or event.correlation_id` selects the first non-empty value, not the first valid UUID. Therefore, an invalid `message_id` hides a valid `correlation_id` and unnecessarily triggers UUID5 fallback.

Please validate them independently in priority order: valid `message_id`, then valid `correlation_id`, then UUID5 fallback.

Vieran (changes requested):

extensions/eda/plugins/event_source/azure_service_bus.py
ttuffin: I have addressed this by generating the UUID based on `queue_name` and `sequence_number`. AFAICT, `sequence_number` should always be assigned by the broker, although looking at the code it is possible for it to return None, in which case messages may still produce identical UUID's. I will leave it to your guidance if its still valid enough to use for seed generation.

extensions/eda/plugins/event_source/azure_event_hub.py
ttuffin: I have added a new function (`_normalize_amqp_id`) to normalize the ID.

extensions/eda/plugins/event_source/azure_service_bus.py
ttuffin: Fixed.

extensions/eda/plugins/event_source/azure_event_hub.py
ttuffin: Updated to validate them in order.

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -22,6 +21,11 @@
   - An ansible-rulebook event source module for receiving events from an Azure service bus.
   - In order to get the service bus and the connection string, refer to
     https://learn.microsoft.com/en-us/azure/service-bus-messaging/service-bus-python-how-to-use-queues?tabs=passwordless
+  - Each event is assigned a valid RFC 4122 UUID under meta.uuid.
+  - If the message_id from Azure Service Bus is already a valid UUID it is used directly.
+  - If the message_id is not a valid UUID, a deterministic UUID5 is generated from
+    queue_name and sequence_number (broker-assigned, unique within the queue) to ensure
+    global uniqueness. The original value is always preserved under meta.message_id.
```
Vieran: The plugin documentation currently says this guarantees "global
uniqueness". Since the seed is unique within the configured queue rather than
across Azure namespaces, could we change that wording to "uniqueness within
the configured queue"?

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -74,6 +78,45 @@
...
+def _process_service_bus_uuid(
+    msg: ServiceBusReceivedMessage,
+    queue_name: str,
+    meta: dict[str, Any],
+) -> None:
+    """Assign a valid UUID to meta["uuid"].
+
+    Uses msg.message_id if it is a valid UUID, otherwise generates a
+    deterministic UUID5 from queue_name and message_id to ensure global
+    uniqueness. The original message_id is always preserved under
```
Vieran: Could you update both descriptions to match the implementation?

Vieran (changes requested):
Thanks for the contribution!

A few small comments related to documentations.

The PR description still lists the Service Bus UUID5 fallback seed as `msg.message_id`, while the implementation now uses `queue_name:sequence_number`.

extensions/eda/plugins/event_source/azure_service_bus.py
ttuffin: Good catch, I've updated it

extensions/eda/plugins/event_source/azure_service_bus.py
ttuffin: Updated.

ttuffin:
> Thanks for the contribution!
> 
> A few small comments related to documentations.
> 
> The PR description still lists the Service Bus UUID5 fallback seed as `msg.message_id`, while the implementation now uses `queue_name:sequence_number`.

@Vieran I have updated the PR description and corrected the errors in the docstrings.

Vieran (approved):
Thanks! I have verified that the pipeline [tests](https://dev.azure.com/azclitools/internal/_build/results?buildId=333734&view=results) passed. Approved this.

Issue #117, opened 2020-05-14
azure_rm_virtualmachine_scaleset >> muti-nic support ?


nergalex:
##### SUMMARY
Support a multi-nic configuration in the module azure_rm_virtualmachine_scaleset
Currently, only one NIC, one LB can be set.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
Support a multi-nic configuration in the module azure_rm_virtualmachine_scaleset

##### ADDITIONAL INFORMATION

Support requirements encountered in customers: one NIC for management plane and one NIC for dataplane

```yaml

```

Fred-sun:
@nergalex Glad you raised this question, we will investigate and add this feature. Thank you!

PR #2300, opened 2026-07-20
Fix pre-existing bugs and clean up azure_rm_virtualmachinescaleset

zunyangc:
##### SUMMARY

Consolidated cleanup of pre-existing bugs in `azure_rm_virtualmachinescaleset` discovered during the audit for #117 (multi-NIC VMSS support). Landing these fixes first keeps the #117 diff focused purely on the multi-NIC feature.

Fixes latent bugs across four categories:
- Field-name typos that caused user input to be silently ignored (`secure_boot_nabled`, `update_security_profle`, `cummunity_gallery_image_id`, `sale in policy`, and the `securityProfile` camelCase key lookup).
- Argspec defaults (`capacity=1`, `single_placement_group=False`, `os_disk_caching='ReadOnly'`, `orchestration_mode='Flexible'`, `platform_fault_domain_count=1`) that silently reset live VMSS state on unrelated updates — a real hazard for autoscaled or Spot production VMSSes.
- Diff blocks that reported `changed=True` but never assigned the new value to the SDK model — the API call sent the old value and state never actually changed (`upgrade_policy`, `security_profile`).
- A `data_disks` diff that only compared list length, so changing disk size on an existing disk was silently ignored, plus a `KeyError` on update when the user omitted optional fields.
- Missing input validation (`subnet_name` required on create) that surfaced as an unhelpful `AttributeError: 'NoneType' object has no attribute 'id'`.
- `max_price = -1` (argspec default sentinel) treated as a real value in diff, causing Spot VMSSes with a real `max_price` to false-fail "not updatable" on tag-only updates.
- Image dict identifier priority contradicted the Azure REST spec (`name` was preferred over `id`).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_virtualmachinescaleset.py
tests/integration/targets/azure_rm_virtualmachinescaleset/tasks/main.yml

##### ADDITIONAL INFORMATION

**This PR contains behavior changes. It should target a major release (next appropriate SemVer bump), not a minor.**

**Potentially breaking — argspec defaults removed**

Removing these argspec defaults changes what gets sent to Azure on create when the user omits the option. On update, the module now inherits the live value (this is what makes the fix worthwhile).

| Option | Old default | New behavior |
|---|---|---|
| `capacity` | `1` | No default; Azure default applies on create; live value preserved on update |
| `single_placement_group` | `False` | No default; Azure default (`true` for Uniform) applies on create; live value preserved on update |
| `orchestration_mode` | `'Flexible'` | No default; Azure default (`Uniform`) applies on create; immutable on update |
| `platform_fault_domain_count` | `1` | No default; Azure default applies on create; immutable on update |
| `os_disk_caching` | `'ReadOnly'` | No default; Azure default applies on create; live value preserved on update |

Playbooks that omit these options on create will now receive Azure's own defaults instead of the module's. In most cases the outcome is the same, but `orchestration_mode` and `single_placement_group` may differ. Users who rely on any of the above defaults should set them explicitly in their playbooks.

**Behavior change — immutable field errors are now surfaced**

- `zones` change on update now fails with a clear "not updatable" error. Previously it was a silent no-op. Playbooks that were "successfully" doing nothing on a `zones` change will now error.

**Behavior change — image dict priority reordered**

When the `image` dict contains multiple identifier keys, the priority is now `id → community_gallery_image_id → marketplace (publisher/offer/sku/version) → name`. Previously the module preferred `name` over `id`, contradicting the Azure REST spec. Playbooks passing both `name` and `id` will now use `id`.

**Refactor and dead-code cleanup**

- Extracted `build_data_disks()` and `build_security_profile()` helpers to deduplicate ~40 LOC between the create and update paths.
- Extracted `data_disks_differ()` with two normalizer helpers to keep the diff block readable, using `default_compare` from `azure_rm_common_ext`.
- Removed dead `get_virtual_network()` method (only caller was commented out).
- Removed commented-out virtual network lookup cruft.
- Removed duplicate `ResourceNotFoundError` import.

**Integration tests added (~385 LOC)**

- Create VMSS with `security_profile` + `scale_in_policy` and assert live state; idempotence re-run.
- Update without specifying `orchestration_mode` (regression for argspec-default reset).
- Update `data_disks` with minimal fields (regression for KeyError).
- Update `security_profile.uefi_settings.v_tpm_enabled` (regression for typo).
- Preserve-live-values baseline: create with distinctive values, tag-only update, read back and assert `capacity`, `overprovision`, `single_placement_group`, `vm_size`, `os_disk_caching` all unchanged.
- Update `upgrade_policy` Manual → Automatic and assert live-side flip.
- Update `data_disks` `disk_size_gb` 64 → 128 and assert live-side resize.
- Create Spot VMSS with `max_price=0.5`, then tag-only update — assert no false "not updatable" failure.
- Create without `subnet_name` — assert `failed` with clear message.

**Testing**

- `ansible-test sanity --python 3.10` on the module and integration test target passes clean.
- No CHANGELOG edits (per release-time convention).
- No new dependencies.

**Follow-up**

- Prerequisite for #117 (multi-NIC VMSS support) — a follow-up PR will build on this cleanup.

### DIFF
--- a/plugins/modules/azure_rm_virtualmachinescaleset.py
+++ b/plugins/modules/azure_rm_virtualmachinescaleset.py
@@ -1094,8 +1094,6 @@ class AzureRMVirtualMachineScaleSet(AzureRMModuleBaseExt):
                     differences.append('Capacity')
                     changed = True
                     vmss_dict['sku']['capacity'] = self.capacity
-                elif self.capacity is None:
-                    self.capacity = vmss_dict['sku']['capacity']
 
                 if self.vm_size and \
                    self.vm_size != vmss_dict['sku']['name']:
@@ -1144,8 +1142,6 @@ class AzureRMVirtualMachineScaleSet(AzureRMModuleBaseExt):
                    bool(self.single_placement_group) != bool(vmss_dict['single_placement_group']):
                     differences.append('single_placement_group')
                     changed = True
-                elif self.single_placement_group is None:
-                    self.single_placement_group = vmss_dict['single_placement_group']
 
                 vmss_dict['zones'] = [int(i) for i in vmss_dict['zones']] if 'zones' in vmss_dict and vmss_dict['zones'] else None
                 if self.zones is not None and self.zones != vmss_dict['zones']:
@@ -1467,7 +1463,10 @@ class AzureRMVirtualMachineScaleSet(AzureRMModuleBaseExt):
                     vmss_resource.tags = self.tags
 
                     if self.upgrade_policy:
-                        vmss_resource.upgrade_policy = self.compute_models.UpgradePolicy(mode=self.upgrade_policy)
+                        if vmss_resource.upgrade_policy is not None:
+                            vmss_resource.upgrade_policy.mode = self.upgrade_policy
+                        else:
+                            vmss_resource.upgrade_policy = self.compute_models.UpgradePolicy(mode=self.upgrade_policy)
 
                     if self.custom_data and vmss_resource.virtual_machine_profile.os_profile is not None:
                         vmss_resource.virtual_machine_profile.os_profile.custom_data = self.custom_data
--- a/plugins/modules/azure_rm_virtualmachinescaleset.py
+++ b/plugins/modules/azure_rm_virtualmachinescaleset.py
@@ -1094,7 +1094,7 @@ class AzureRMVirtualMachineScaleSet(AzureRMModuleBaseExt):
                     differences.append('Capacity')
                     changed = True
                     vmss_dict['sku']['capacity'] = self.capacity
-                else:
+                elif self.capacity is None:
                     self.capacity = vmss_dict['sku']['capacity']
 
                 if self.vm_size and \
--- a/plugins/modules/azure_rm_virtualmachinescaleset.py
+++ b/plugins/modules/azure_rm_virtualmachinescaleset.py
@@ -58,7 +58,8 @@ options:
     capacity:
         description:
             - Capacity of VMSS.
-        default: 1
+            - When not set on create, Azure defaults to C(1).
+            - When not set on update, the existing capacity is preserved (safe for autoscaled VMSS).
         type: int
     tier:
         description:
@@ -140,11 +141,12 @@ options:
     os_disk_caching:
         description:
             - Type of OS disk caching.
+            - When not set on create, Azure defaults to C(ReadOnly).
+            - When not set on update, the existing value is preserved.
         type: str
         choices:
             - ReadOnly
             - ReadWrite
-        default: ReadOnly
         aliases:
             - disk_caching
     os_disk_size_gb:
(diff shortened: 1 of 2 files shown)


zunyangc:
Hi @Vieran, pipeline [tests](https://dev.azure.com/azclitools/internal/_build/results?buildId=331636&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_virtualmachinescaleset.py
```diff
@@ -1087,12 +1088,14 @@ def exec_module(self, **kwargs):
...
 
-                if self.capacity and \
+                if self.capacity is not None and \
                    self.capacity != vmss_dict['sku']['capacity']:
                     self.log('CHANGED: virtual machine scale set {0} - Capacity'.format(self.name))
                     differences.append('Capacity')
                     changed = True
                     vmss_dict['sku']['capacity'] = self.capacity
+                elif self.capacity is None:
+                    self.capacity = vmss_dict['sku']['capacity']
```
Vieran: When `capacity` is omitted, this line stores the value returned by the initial GET in `self.capacity`. The update path performs another GET at L1456, but L1459-L1460 then overwrite that newly fetched resource with the earlier capacity. If an autoscaler changes the capacity between those two GETs, an unrelated update, such as a tag change, will submit the stale value and revert the autoscaler change.

Please leave `self.capacity` as `None` when the option is omitted. The resource returned by the second GET already contains the latest live capacity, and the existing conditional assignment at L1459 will then preserve it.

plugins/modules/azure_rm_virtualmachinescaleset.py
```diff
@@ -1489,15 +1454,21 @@ def exec_module(self, **kwargs):
...
+                    if self.vm_size is not None:
+                        vmss_resource.sku.name = self.vm_size
+                    if self.overprovision is not None:
+                        vmss_resource.overprovision = self.overprovision
+                    if self.single_placement_group is not None:
+                        vmss_resource.single_placement_group = self.single_placement_group
                     vmss_resource.tags = self.tags
 
+                    if self.upgrade_policy:
+                        vmss_resource.upgrade_policy = self.compute_models.UpgradePolicy(mode=self.upgrade_policy)
```
Vieran: I understand this assignment is intended to ensure the updated `mode` is sent to Azure.

The object returned by `get_vmss()` may contain `rolling_upgrade_policy` or `automatic_os_upgrade_policy`, while the new object only retains `mode`. Would updating `vmss_resource.upgrade_policy.mode` in place be safer, or are the nested policies expected to be reset when the mode changes?

Vieran (changes requested):
Thanks for the contribution! Mostly LGTM. I left comments for some small issues.

zunyangc:
Hi @Vieran, updated as per requested. [Test](https://dev.azure.com/azclitools/internal/_build/results?buildId=333977&view=results) passed. Ready to merge, thanks!

Vieran (commented):
LGTM. But this code contains breaking change, should we merge it directly into the dev branch? Or we should create a separated branch for it?

Vieran (approved):

PR #2311, opened 2026-07-27
emit signals from arc extensions module

stefanoochoa:
##### SUMMARY

Log success/failure signals for the azure_rm_arcmachineextensions module

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME
azure_rm_arcmachineextensions

##### ADDITIONAL INFORMATION

Tested simulating success/failure scenarios and confirmed signals showed up service side.

### DIFF
--- a/plugins/modules/azure_rm_arcmachineextensions.py
+++ b/plugins/modules/azure_rm_arcmachineextensions.py
@@ -383,9 +383,9 @@ class AzureRMarcmachineextensions(AzureRMModuleBaseExt):
         # Best-effort signal about the applied change (never on check mode / no-op).
         if self.results['changed'] and not self.check_mode:
             if self.state == 'present':
-                self._emit_signal('info', "extension '{0}' on machine '{1}' applied".format(self.name, self.machine_name), 'extensionUpdate')
+                self._emit_signal('info', "extension '{0}' on machine '{1}' applied".format(self.name, self.machine_name), 'ansibleExtensionUpdate')
             elif self.state == 'absent':
-                self._emit_signal('info', "extension '{0}' on machine '{1}' deleted".format(self.name, self.machine_name), 'extensionDelete')
+                self._emit_signal('info', "extension '{0}' on machine '{1}' deleted".format(self.name, self.machine_name), 'ansibleExtensionDelete')
 
         return self.results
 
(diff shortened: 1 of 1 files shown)


zunyangc (approved):
LGTM, thanks!

zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=333372&view=results) passed. Ready to merge, thanks!

Klaas-:
Hi Stefan,
I think think this may not be the best idea, if you're adding stuff like this to all modules you'll duplicate a lot of code. Talk with the main maintainers of this project, decide on how to do telemetry and centralize it in module utils. Maybe it's an even better idea to move it into azure python sdk directly?

Now that for the technical part of this change -- now the "high level" part -- I don't think you need telemetry here at all. This module does webcalls to azure-api and it identifies via User Agent. So you have telemetry on the server side already. If you want to distinguish between direkt python sdk users and ansible users then the solution may be as easy as to overwrite user agent on python sdk usage or something like that.


Greetings
Klaas

stefanoochoa:
Hey @Klaas- ! 
> if you're adding stuff like this to all modules you'll duplicate a lot of code

The telemetry is scoped just to the Arc extension module & Arc onboarding role. If that wasn't the case, then I'd agree with your recommended approach (centralizing in module utils).

> This module does webcalls to azure-api and it identifies via User Agent. So you have telemetry on the server side already. If you want to distinguish between direkt python sdk users and ansible users then the solution may be as easy as to overwrite user agent on python sdk usage or something like that.

We want telemetry specifically on the module usage. We can't rely on User Agent alone, as that does not differentiate between users that are actually using the module vs some other custom ansible approach.

LMK if you have any other concerns; I appreciate the feedback 🙂

Klaas-:
https://learn.microsoft.com/en-us/python/api/azure-core/azure.core.pipeline.policies.useragentpolicy?view=azure-python you can just set user agent, I am guessing setting it to something that includes the initiating module name would be an easy solution to make server side telemetry easy for all azure ansible modules

stefanoochoa:
> https://learn.microsoft.com/en-us/python/api/azure-core/azure.core.pipeline.policies.useragentpolicy?view=azure-python you can just set user agent, I am guessing setting it to something that includes the initiating module name would be an easy solution to make server side telemetry easy for all azure ansible modules

That still would not give us as much insight/coverage as we want here. That would only allow us to track the result of operations which reach Azure; it does not give visibility into failures in the module itself.

PR #2313, opened 2026-07-28
feat: Add checkpoint store, service principal auth, and topic/subscription

mkanoor:
##### SUMMARY
Enhance EDA Source Plugins
##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
* azure.azcollection.azure_service_bus
* azure.azcollection.azure_event_hub

**Azure Event Hub:**
- Add persistent checkpoint storage via Azure Blob Storage (azure_storage_account_name, azure_checkpoint_container_name)
- When checkpoint store is configured, starting_position is omitted from receive() to let the SDK correctly resume from checkpoints
- Add azure_max_wait_time parameter for load balancing interval control
- Change default starting_position from "-1" to "@latest"
- Coerce numeric starting_position strings to int for backward compatibility with existing rulebooks passing "-1" or other offsets

**Azure Service Bus:**
- Add service principal authentication (azure_tenant_id, azure_client_id, azure_client_secret, azure_namespace) as an alternative to conn_str
- Add topic/subscription support via topic_name and subscription_name
- Add dead-letter handling on feedback timeout and processing errors
- Validate mutually exclusive queue vs topic/subscription configuration

New Python package dependencies:
- azure-eventhub-checkpointstoreblob-aio (for persistent checkpoint storage)
- azure-identity (for service principal authentication via ClientSecretCredential)

Tests:
- Add tests for checkpoint store validation, creation, and receive behavior
- Add tests for starting_position type coercion (numeric vs string)
- Add tests for max_wait_time configuration
- Add tests for _create_service_bus_client with both auth methods
- Add tests for topic/subscription receiver and validation
- Add tests for dead-letter on feedback timeout
- Update existing tests for new defaults and refactored internals

### DIFF
--- a/extensions/eda/plugins/event_source/azure_event_hub.py
+++ b/extensions/eda/plugins/event_source/azure_event_hub.py
@@ -15,6 +15,7 @@ from typing import Any
 
 from azure.eventhub import EventData
 from azure.eventhub.aio import EventHubConsumerClient, PartitionContext
+from azure.eventhub.extensions.checkpointstoreblobaio import BlobCheckpointStore
 from azure.identity.aio import ClientSecretCredential
 
 DOCUMENTATION = r"""
@@ -60,14 +61,39 @@ options:
     required: true
   azure_starting_position:
     description:
-      - The starting position
+      - The starting position valid values are -1, @latest, integer position
+        or datetime. https://azuresdkdocs.z19.web.core.windows.net/python/azure-eventhub/latest/azure.eventhub.html
     type: str
-    default: "-1"
+    default: "@latest"
+
   azure_consumer_group:
     description:
       - The name of the consumer group
     type: str
     default: "$Default"
+  azure_storage_account_name:
+    description:
+      - The Azure Storage Account name for persisting checkpoints
+      - Uses the same service principal credentials as Event Hub
+      - If not provided, checkpoints will only be stored in-memory
+      - Example "mystorageaccount" (not the full URL)
+    type: str
+    required: false
+  azure_checkpoint_container_name:
+    description:
+      - The name of the blob container to store checkpoints
+      - Required if azure_storage_account_name is provided
+    type: str
+    required: false
+  azure_max_wait_time:
+    description:
+      - Maximum time in seconds to wait for events before checking for partition rebalancing
+      - Lower values (10-30) = faster failover detection but more Azure API calls
+      - Higher values (60-120) = slower failover but fewer API calls and lower costs
+      - Only relevant when using load balancing (multiple consumers with same consumer group)
+    type: int
+    default: 60
+    required: false
   feedback:
     type: bool
     default: false
@@ -101,7 +127,10 @@ EXAMPLES = r"""
     "azure_client_secret": "your_client_secret"
     "azure_namespace": "example.servicebus.windows.net"
     "azure_event_hub_name": "your_hub_name"
-    "azure_starting_position": "-1"
+    "azure_consumer_group": "$Default"
+    "azure_starting_position": "@latest"
+    "azure_storage_account_name": "mystorageaccount"
+    "azure_checkpoint_container_name": "eventhub-checkpoints"
 """
 
 logger = logging.getLogger()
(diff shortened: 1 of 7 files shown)


mkanoor:
will wait for PR #2299 to get merged first

extensions/eda/plugins/event_source/azure_event_hub.py
```diff
@@ -60,14 +61,39 @@
     required: true
   azure_starting_position:
     description:
-      - The starting position
+      - The starting position valid values are -1, @latest, integer position
+        or datetime. https://azuresdkdocs.z19.web.core.windows.net/python/azure-eventhub/latest/azure.eventhub.html
     type: str
-    default: "-1"
+    default: "@latest"
```
zunyangc, line 67: (Breaking Change) This changes behavior for existing rulebooks that omit `azure_starting_position`. Is this intentional?

zunyangc (changes requested):
Hi @mkanoor thanks for this PR, LGTM, just with a small comment.

extensions/eda/plugins/event_source/azure_event_hub.py
mkanoor: @zunyangc It is intentional. The earlier default does not make sense in production and doesn't match the defaults provided by azure sdk.
```
Also supported are the values “-1” for receiving from the beginning of the stream, and “@latest” for receiving only new events. Default value is “@latest”.
```
For the Event Driven Ansible usecase the default of -1 would mean always read from the beginning which will cause the same events to be fired everytime the process starts resulting in repeating all the actions for every event.

zunyangc (approved):
LGTM. thanks!

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -126,54 +190,199 @@ def _process_service_bus_uuid(
...
+        async with receiver:
+            destination_name = args.get("queue_name") or args.get(
+                "azure_topic_name", "unknown",
+            )
+            async for msg in receiver:
+                try:
+                    await _process_message(
+                        msg, receiver, queue, feedback, destination_name,
+                    )
+                except Exception as e:
```
zunyangc, line 365: `_process_message` already calls `receiver.dead_letter_message(...)` on `asyncio.TimeoutError` and re-raises. This outer except Exception then catches that same `TimeoutError` and attempts a second `dead_letter_message(...)`, which fails with `MessageAlreadySettled` / `MessageLockLostError` and is swallowed by the nested `except Exception`:.

extensions/eda/plugins/event_source/azure_service_bus.py
```diff
@@ -126,54 +190,199 @@ def _process_service_bus_uuid(
...
+        ServiceBusClient instance.
+
+    Raises:
+        ValueError: If neither connection string nor service principal credentials are provided.
+
+    """
+    logging_enable = bool(args.get("logging_enable", True))
+
+    # Option 1: Connection string authentication (backward compatible)
+    if "conn_str" in args:
```
zunyangc: Shall we use `if args.get("conn_str"):` instead?

zunyangc (changes requested):
Small change requests. Thanks!

ramensocode:
@mkanoor  Any update on this PR progress ?

zunyangc (approved):

extensions/eda/plugins/event_source/azure_event_hub.py
```diff
@@ -267,14 +319,70 @@ async def on_event(
...
             eventhub_name=self.event_hub_name,
             consumer_group=self.consumer_group,
             credential=self.credential,
+            checkpoint_store=checkpoint_store,
         )
+
+        # Configure receive arguments
+        receiver_args = {
+            "on_event": self.on_event,
+            "max_wait_time": self.max_wait_time,  # Controls load balancing interval
```
Vieran, line 370: The `max_wait_time` controls idle `on_event(None)` callbacks, not partition rebalancing. Could we pass this value as `load_balancing_interval` to `EventHubConsumerClient` instead, so the option works as documented?

Vieran (changes requested):
Thanks for opening this! Mostly LGTM. One comment left.

PR #2314, opened 2026-07-29
Add some support for addons

desand01:
##### SUMMARY

This PR enhances the `azure_rm_aks` module by improving support for AKS managed cluster add-ons.

Changes included:
- Added support for the `use_aad_auth` option of the **monitoring** add-on (`useAADAuth` in the Azure API).
- Added support for the `enable_secret_rotation` and `rotation_poll_interval` options of the **azure_keyvault_secrets_provider** add-on.
- Refactored the internal add-on configuration model to support optional and typed add-on parameters.
- Add-on serialization/deserialization preserve Python types (for example, Azure `"true"`/`"false"` values are converted back to Python booleans).
- Ignoring unspecified optional add-on parameters during comparison with the current Azure state.

The PR also includes a output compatibility change:

Before:

```yaml
addon:
  httpApplicationRouting:
    enabled: false
  azureKeyvaultSecretsProvider:
    enabled: true
```

After:

```yaml
addon:
  http_application_routing:
    enabled: false
  azure_keyvault_secrets_provider:
    enabled: true
```

The returned add-on names are now consistent with the Ansible module input parameters.

Additionally, the following sanity ignore entry was added because `validate-modules`
incorrectly detects `azure_keyvault_secrets_provider` and
`enable_secret_rotation` as secret values due to their names:

```text
plugins/modules/azure_rm_aks.py validate-modules:no-log-needed
```

##### ISSUE TYPE

- Feature Pull Request

##### COMPONENT NAME

azure_rm_aks

##### ADDITIONAL INFORMATION

This change is backward compatible except for the normalization of the
returned add-on names in the module output.

The implementation keeps the existing Azure API behavior while making the module
easier to extend as new AKS add-on configuration options become available.

The following add-on options are now supported:

```yaml
addon:
  monitoring:
    enabled: true
    log_analytics_workspace_resource_id: /subscriptions/.../workspaces/myworkspace
    use_aad_auth: true

  azure_keyvault_secrets_provider:
    enabled: true
    enable_secret_rotation: true
    rotation_poll_interval: 2m
```


tests/sanity/ignore-2.21.txt
```diff
@@ -6,3 +6,4 @@ plugins/modules/azure_rm_servicebussaspolicy.py validate-modules:bad-return-valu
 plugins/modules/azure_rm_tags_info.py validate-modules:bad-return-value-key
 plugins/modules/azure_rm_vmsku_info.py validate-modules:bad-return-value-key
 plugins/module_utils/azure_rm_common.py pylint:ansible-bad-function
+plugins/modules/azure_rm_aks.py validate-modules:no-log-needed
```
zunyangc: Can we use `no_log=False` instead of adding this entry into ignore?

zunyangc (changes requested):
Thanks for this PR, mostly LGTM, I left a small comment.

tests/sanity/ignore-2.21.txt
desand01: Yes, that's much better. done

plugins/modules/azure_rm_aks.py
```diff
@@ -491,6 +491,22 @@
                             - Enabled or disabled the Azure Key Vault provider in the AKS cluster.
                         type: bool
                         default: true
+                    enable_secret_rotation:
+                        description:
+                            - Enable Key Vault Secret rotation.
+                        type: bool
+                        version_added: '3.22.0'
```
zunyangc:
```suggestion
                        version_added: '4.0.0'
```

Since we plan to release a major version end of the month. Thanks!

plugins/modules/azure_rm_aks.py
```diff
@@ -491,6 +491,22 @@
...
+                        required: false
+                        aliases:
+                            - enableSecretRotation
+                    rotation_poll_interval:
+                        description:
+                            - Polling interval used for rotation.
+                        type: str
+                        version_added: '3.22.0'
```
zunyangc:
```suggestion
                        version_added: '4.0.0'
```

plugins/modules/azure_rm_aks.py
```diff
@@ -511,6 +527,15 @@
                         required: true
                         aliases:
                             - logAnalyticsWorkspaceResourceID
+                    use_aad_auth:
+                        description:
+                            - Enables managed identity authentication between the cluster monitoring agent
+                              and the Log Analytics workspace.
+                        type: bool
+                        version_added: '3.22.0'
```
zunyangc:
```suggestion
                        version_added: '4.0.0'
```

zunyangc (changes requested):
Small changes required, thanks!

plugins/modules/azure_rm_aks.py
```diff
@@ -1584,14 +1686,14 @@ def compare_addon(origin, patch, config):
                             return False
                         config = config or dict()
                         for key in config.keys():
-                            if origin.get(config[key]) != patch.get(key):
-                                return False
+                            if origin.get(key) != patch.get(key):
```
zunyangc: This loop exits on the First differing key. If that key happens to be unspecified by the user (`patch.get(key) is None`), the function returns True and skips every remaining key, including keys the user did set to a new value. 

Concretely, if a caller changes only `rotation_poll_interval` and leaves `enable_secret_rotation` unset, and Azure returned `enable_secret_rotation=True`, iteration hits `enable_secret_rotation` first, sees `patch.get(...) is None`, and returns "equal" without ever comparing `rotation_poll_interval`. 

Replace the `return patch.get(key) is None` with `if patch.get(key) is None: continue` and return False only when user supplied a differing value; return True after the loop.

plugins/modules/azure_rm_aks.py
```diff
@@ -1040,13 +1065,49 @@ def create_api_server_access_profile_dict(api_server):
...
+                }
+                for ansible_key, config in (spec.get('config') or {}).items()
+            }
+        )
+        for ansible_name, spec in ADDONS.items()
+    }
+
+    for azure_name, profile in addon.items():
+        if azure_name not in reverse_addons:
+            continue
```
zunyangc, line 1099: This will drop any add-on Azure returns that is not in the module's `ADDONS`.

zunyangc (changes requested):
Hi @desand01, apologies for the follow up changes request, could you please check on these two? Thanks!

plugins/modules/azure_rm_aks.py
desand01: Ok, the fix will forward the configuration received from Azure for unsupported add-ons.

zunyangc (approved):
LGTM thanks!

zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=337958&view=results) passed. Ready to merge, thanks!

Vieran (approved):

PR #2321, opened 2026-08-03
4.0.0 Removal: `azure_rm_sqldatabase` remove deprecated `edition` option

zunyangc:
##### SUMMARY
Part of the 4.0.0 deprecation removal sweep. Removes the long-deprecated `edition` option from `azure_rm_sqldatabase`, in favor of the SDK-native `sku` option. `edition` was deprecated in **2.11 (2021)** with `sku` shipped as the direct replacement — internally the module already converted `edition` to `sku` via a hardcoded mapping table (`get_sku_name`).

`sku` matches the Azure Python SDK (`azure-mgmt-sql`) `Database.sku` field and the Azure REST API `Databases - Create Or Update` shape one-to-one.

While auditing the module against the SDK model, a pre-existing latent bug was found on `recovery_services_recovery_point_resource_id`: the SDK field name is [`recovery_services_recovery_point_id`](https://github.com/Azure/azure-sdk-for-python/blob/azure-mgmt-sql_4.0.0b19/sdk/sql/azure-mgmt-sql/azure/mgmt/sql/models/_models_py3.py#L1167) (no `_resource_`), and the SDK constructor silently dropped the unknown kwarg. Users passing this option with `create_mode=restore_long_term_retention_backup` were getting a no-op on that field. Since this is a hard rename that only belongs in a major bump, it's included here:
- Renamed the option to `recovery_services_recovery_point_id` (matches SDK).
- Added `required_if` for `create_mode=restore_long_term_retention_backup` so the module fails fast.

##### ISSUE TYPE
- Removal Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_sqldatabase.py

### DIFF
--- a/plugins/modules/azure_rm_sqldatabase.py
+++ b/plugins/modules/azure_rm_sqldatabase.py
@@ -50,7 +50,7 @@ options:
             - C(recovery), Creates a database by restoring a geo-replicated backup.
             - C(restore), Creates a database by restoring a backup of a deleted database.
             - C(restore_long_term_retention_backup), Creates a database by restoring from a long term retention vault.
-            - C(copy), C(non_readable_secondary), C(online_secondary) and C(restore_long_term_retention_backup) are not supported for C(data_warehouse) edition.
+            - C(copy), C(non_readable_secondary), C(online_secondary) and C(restore_long_term_retention_backup) are not supported for Data Warehouse SKUs.
         type: str
         choices:
             - 'copy'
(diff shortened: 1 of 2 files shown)


zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=334865&view=results) passed. Ready to merge, thanks!

Vieran (approved):

PR #2328, opened 2026-08-05
4.0.0 Removal: `azure_rm_containerinstance` remove `ports` option

zunyangc:
##### SUMMARY
The top-level 'ports' option was a no-op since the module was created: `argspec` was declared but `self.ports` is never read at runtime. Port exposure is entirely driven by the per-container 'ports' inside each 'containers:' entry, and the union of those ports is what the module already surfaces on `IpAddress`. This aligns the module with the Azure REST API contract (`Container.ports` is the only ports field; container groups have no such field of their own).

##### ISSUE TYPE
- Removal Pull Request

##### COMPONENT NAME
plugins/modules/azure_rm_containerinstance.py

### DIFF
--- a/plugins/modules/azure_rm_containerinstance.py
+++ b/plugins/modules/azure_rm_containerinstance.py
@@ -59,13 +59,6 @@ options:
         description:
             - The Dns name label for the IP.
         type: str
-    ports:
-        description:
-            - List of ports exposed within the container group.
-            - This option is deprecated, using I(ports) under I(containers)".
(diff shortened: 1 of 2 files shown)


zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=335385&view=results) passed. Ready to merge, thanks!

Vieran (approved):

PR #2331, opened 2026-08-06
Fix 2322: Update `attach_caching` on managed disk when managed_by is not resupplied

### LINKED ISSUE #2322: azure_rm_manageddisk does not update attach_caching, if the size does not change
##### SUMMARY
attach_caching of a manageddisk is not updated for an existing disk.
Task state is just "ok".

Creating a new disk with attach_caching of '', read_write, read_only - works just fine.
Updating a disk's attach_caching to '', read_write, read_only - task is ok without changing anything.

##### ISSUE TYPE
- Bug Report

##### COMPONENT NAME
azure_rm_manageddisk

##### ANSIBLE VERSION

```paste below
2.16
```

##### COLLECTION VERSION

```paste below
3.21.0
```

##### CONFIGURATION

```paste below

```

##### OS / ENVIRONMENT

RHEL 9

##### STEPS TO REPRODUCE

```yaml

```

##### ACTUAL RESULTS

```paste below

```

zunyangc:
##### SUMMARY
Fix #2322.
`azure_rm_manageddisk` previously reported `ok` and applied no change when a caller updated only `attach_caching` on an existing attached disk without re-specifying `managed_by`, because the caching diff / detach-reattach path was gated behind `self.managed_by`. This patch backfills `self.managed_by` from the current disk when the user updates caching only, so the existing caching-diff path runs and the VM's `DataDisk.caching` is actually updated.

##### ISSUE TYPE
- [ ] New Module Pull Request
- [x] Bug Fix Pull Request
- [ ] Test Fix Pull Request
- [ ] Migration Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_manageddisk.py

##### ADDITIONAL INFORMATION
- On an existing attached managed disk, calling the module with only `attach_caching` (no `managed_by`) now correctly detects the caching drift and triggers the detach/reattach path that updates `DataDisk.caching` on the VM.
- Scope is intentionally narrow: the backfill fires only when `attach_caching is not None and managed_by is None`, so behavior for every other invocation (create, size change, tag change, explicit `managed_by`) is unchanged.
- Caching is a VM-side property (`VirtualMachine.storage_profile.data_disks[*].caching`), so the fix reuses the module's existing detach/reattach flow rather than introducing a disk-level PATCH.

##### REFERENCE
- https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/compute/azure-mgmt-compute/azure/mgmt/compute/v2024_07_01/models/_models_py3.py
- https://github.com/Azure/azure-rest-api-specs/tree/main/specification/compute/resource-manager/Microsoft.Compute/ComputeRP/stable/2024-07-01

### DIFF
--- a/plugins/modules/azure_rm_manageddisk.py
+++ b/plugins/modules/azure_rm_manageddisk.py
@@ -882,6 +882,10 @@ class AzureRMManagedDisk(AzureRMModuleBase):
                 self.image_reference = disk_instance.get('image_reference')
             if self.gallery_image_reference is None:
                 self.gallery_image_reference = disk_instance.get('gallery_image_reference')
+            if self.attach_caching is not None and self.managed_by is None:
+                current_managed_by = parse_resource_id(disk_instance.get('managed_by', '')).get('name') if disk_instance.get('managed_by') else None
+                if current_managed_by:
+                    self.managed_by = current_managed_by
 
         result = disk_instance
 
--- a/tests/integration/targets/azure_rm_manageddisk/tasks/main.yml
+++ b/tests/integration/targets/azure_rm_manageddisk/tasks/main.yml
@@ -418,6 +418,100 @@
           - output.state.tags.testtag2 == 'TestValue2'
           - output.state.tags.TestTag3 == 'TestValue3'
 
+    - name: Create virtual network for attach_caching test
+      azure.azcollection.azure_rm_virtualnetwork:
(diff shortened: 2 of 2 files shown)


zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=335849&view=results) passed. Ready to merge, thanks!

Vieran (approved):

PR #2345, opened 2026-08-20
Add support for storage profile

desand01:
##### SUMMARY
Add support for configuring the AKS `storage_profile` in the `azure_rm_aks` module.

This includes support for the following storage components:
- Azure Disk CSI Driver
- Azure File CSI Driver
- Snapshot Controller
- Azure Blob CSI Driver

The storage profile is also returned in the module output.

##### ISSUE TYPE
- Feature Pull Request

##### COMPONENT NAME
azure_rm_aks

##### ADDITIONAL INFORMATION
The following configuration is now supported:

```yaml
storage_profile:
  disk_csi_driver:
    enabled: true
  file_csi_driver:
    enabled: true
  snapshot_controller:
    enabled: true
  blob_csi_driver:
    enabled: true
```

### DIFF
--- a/plugins/modules/azure_rm_aks.py
+++ b/plugins/modules/azure_rm_aks.py
@@ -682,6 +682,53 @@ options:
                     - Unmanaged
                     - SecurityPatch
                     - NodeImage
+    storage_profile:
+        description:
+            - Storage profile for the AKS cluster.
+            - Defines the CSI drivers and snapshot controller enabled on the cluster.
+        type: dict
+        version_added: '4.0.0'
+        suboptions:
+            disk_csi_driver:
+                description:
+                    - Azure Disk CSI driver configuration.
+                type: dict
+                suboptions:
+                    enabled:
+                        description:
+                            - Whether the Azure Disk CSI driver is enabled.
+                        type: bool
+                        required: false
+            file_csi_driver:
+                description:
+                    - Azure File CSI driver configuration.
+                type: dict
+                suboptions:
+                    enabled:
+                        description:
+                            - Whether the Azure File CSI driver is enabled.
+                        type: bool
+                        required: false
+            snapshot_controller:
+                description:
+                    - Snapshot controller configuration.
+                type: dict
+                suboptions:
+                    enabled:
+                        description:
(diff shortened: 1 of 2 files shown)


plugins/modules/azure_rm_aks.py
```diff
@@ -1823,6 +1900,11 @@ def compare_addon(origin, patch, config):
                         to_be_updated = True
                         # self.module.warn("windows_profile.admin_username cannot be updated")
 
+                    if not self.default_compare({}, self.storage_profile, response['storage_profile'], '', dict(compare=[])):
```
zunyangc, line 1903: Could we wrap this into `ManagedClusterStorageProfile(**self.storage_profile) if self.storage_profile else None`, the SDK types the field as `Optional[ManagedClusterStorageProfile]`.

zunyangc (changes requested):
Hi @desand01, small change request, thanks!

plugins/modules/azure_rm_aks.py
desand01: Hi, thanks for your review. Do you refer to the creation of the ManagedCluster model in create_update_aks line 2053?
I think we need to keep default_compare and to_be_updated = True logic.

Should we replace:
```
storage_profile=self.storage_profile,
```
with:
```
storage_profile=self.managedcluster_models.ManagedClusterStorageProfile(**self.storage_profile) if self.storage_profile else None,
```

zunyangc: Yes, sorry for the confusion, line 2053 in `create_update_aks()`, inside the `ManagedCluster(...)` call, is exactly where I meant. Please keep the `default_compare` / `to_be_updated = True` block at line 1903 unchanged; that's about diff detection and is independent from how we hand the dict to the SDK.

Your proposed one-liner is correct:
```python
storage_profile=self.managedcluster_models.ManagedClusterStorageProfile(**self.storage_profile) if self.storage_profile else None,

desand01: Done

zunyangc (approved):
LGTM, thanks!

Vieran (approved):

PR #2353, opened 2026-08-25
New Module: Add Azure Container Apps modules

zunyangc:
##### SUMMARY
Adds first-class Ansible support for Azure Container Apps by introducing four new modules — `azure_rm_containerappenvironment` and `azure_rm_containerapp` for CRUD, plus their `_info` counterparts — targeting `azure-mgmt-appcontainers` 5.0.0 (default). Users can now create, update, start, stop, and delete container apps and their managed environments without falling back to `azure_rm_resource`. The modules follow the established `azure_rm_keyvault` 14.0.1 hybrid-model pattern: they construct explicit SDK model instances so ARM receives the required `properties`-envelope wire shape, and use `azure.core.serialization.as_attribute_dict` on responses so returned state is snake_case and flat.

##### ISSUE TYPE
- [x] New Module Pull Request
- [ ] Bug Fix Pull Request
- [ ] Test Fix Pull Request
- [ ] Migration Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_containerappenvironment.py
- plugins/modules/azure_rm_containerappenvironment_info.py
- plugins/modules/azure_rm_containerapp.py
- plugins/modules/azure_rm_containerapp_info.py
- plugins/module_utils/azure_rm_common.py
- meta/runtime.yml
- pr-pipelines.yml
- requirements.txt
- tests/integration/targets/azure_rm_containerapp/aliases
- tests/integration/targets/azure_rm_containerapp/meta/main.yml
- tests/integration/targets/azure_rm_containerapp/tasks/main.yml

##### ADDITIONAL INFORMATION
- Managed environment CRUD via `azure_rm_containerappenvironment`: supports VNet configuration, workload profiles, Log Analytics destination, peer authentication (mTLS), peer traffic encryption, public network access, infrastructure resource group, and zone redundancy.
- Container app CRUD via `azure_rm_containerapp`: full `template` (containers and init containers with probes, env, resources, volume mounts), `configuration` (secrets, registries, ingress with traffic weights and custom domains, Dapr), `scale` with polymorphic rules (HTTP, TCP, Azure Queue, custom), managed identity (system and user-assigned), and `status: start` / `status: stop` lifecycle actions.
- Info modules support single-name lookup and resource-group list. `azure_rm_containerapp_info` optionally calls `list_secrets` and registers the returned values with `module.no_log_values` to prevent log leakage.
- Payload construction uses explicit `azure-mgmt-appcontainers` typed model instances (`ContainerApp`, `ManagedEnvironment`, `Configuration`, `Template`, `Ingress`, `Scale`, `Container`, etc.) so ARM receives resource fields nested under `properties` with correct camelCase serialization; mirrors the pattern already established by `azure_rm_keyvault` after its 14.0.1 migration.
- Response normalization uses `azure.core.serialization.as_attribute_dict`, the Azure SDK's official backcompat helper for hybrid/typed models.
- `azure-mgmt-appcontainers==5.0.0` added to `requirements.txt`; `containerapps_client` accessor added to `plugins/module_utils/azure_rm_common.py` with API version being SDK stable default.
- `meta/runtime.yml` registers the four new modules in `action_groups.azure`; `pr-pipelines.yml` gains an `azure_rm_containerapp` matrix entry.
- Integration test target follows the `azure_rm_mysqlflexibleserver` layout: dedicated `{{ resource_group }}-containerapp` RG, `block:` with `always:` force-delete cleanup, and check-mode / create / idempotence / update / info / start-stop / delete coverage with idempotence assertions on each mutating path.

##### REFERENCE
- https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/appcontainers/azure-mgmt-appcontainers
- https://github.com/Azure/azure-rest-api-specs/tree/main/specification/app/resource-manager/Microsoft.App/stable/2024-03-01

### DIFF
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -265,6 +265,7 @@ try:
     from azure.mgmt.mysqlflexibleservers import MySQLManagementClient as MySQLFlexibleManagementClient
     from azure.mgmt.containerregistry import ContainerRegistryManagementClient
     from azure.mgmt.containerinstance import ContainerInstanceManagementClient
+    from azure.mgmt.appcontainers import ContainerAppsAPIClient
     from azure.mgmt.loganalytics import LogAnalyticsManagementClient
     import azure.mgmt.loganalytics.models as LogAnalyticsModels
     from azure.mgmt.automation import AutomationClient
@@ -425,6 +426,7 @@ class AzureRMModuleBase(object):
         self._containerinstance_client = None
         self._containerservice_client = None
         self._managedcluster_client = None
+        self._containerapps_client = None
         self._traffic_manager_management_client = None
         self._monitor_autoscale_settings_client = None
         self._monitor_log_profiles_client = None
@@ -1339,6 +1341,15 @@ class AzureRMModuleBase(object):
                                                                       api_version='2023-05-01')
 
         return self._containerinstance_client
+
+    @property
+    def containerapps_client(self):
+        self.log('Getting Container Apps client')
+        if not self._containerapps_client:
+            self._containerapps_client = self.get_mgmt_svc_client(ContainerAppsAPIClient,
+                                                                  base_url=self._cloud_environment.endpoints.resource_manager,
+                                                                  api_version='2024-03-01')
+        return self._containerapps_client
 
     @property
     def marketplace_client(self):
--- a/plugins/module_utils/azure_rm_common.py
+++ b/plugins/module_utils/azure_rm_common.py
@@ -1347,8 +1347,7 @@ class AzureRMModuleBase(object):
         self.log('Getting Container Apps client')
         if not self._containerapps_client:
             self._containerapps_client = self.get_mgmt_svc_client(ContainerAppsAPIClient,
-                                                                  base_url=self._cloud_environment.endpoints.resource_manager,
-                                                                  api_version='2024-03-01')
+                                                                  base_url=self._cloud_environment.endpoints.resource_manager)
         return self._containerapps_client
 
     @property
--- a/plugins/modules/azure_rm_containerapp.py
+++ b/plugins/modules/azure_rm_containerapp.py
@@ -1014,8 +1014,6 @@ class AzureRMContainerApp(AzureRMModuleBaseExt):
                         changed = True
                         self.to_do = Actions.Update
 
-                # environment id is immutable in Microsoft.App — warn on drift instead of
-                # flipping Update (the PATCH body would silently drop it and misreport changed).
                 desired_env = self.parameters.get('managed_environment_id')
                 current_env = old_response.get('managed_environment_id')
                 if desired_env and current_env and desired_env.lower() != current_env.lower():
@@ -1024,7 +1022,6 @@ class AzureRMContainerApp(AzureRMModuleBaseExt):
                         "Container Apps do not support moving between environments, ignoring the change."
                     )
 
-                # workload profile is patchable; compare case-insensitively (ARM normalizes casing).
                 desired_wp = self.parameters.get('workload_profile_name')
                 current_wp = old_response.get('workload_profile_name')
                 if desired_wp is not None and (current_wp is None or desired_wp.lower() != current_wp.lower()):
--- a/plugins/modules/azure_rm_containerapp.py
+++ b/plugins/modules/azure_rm_containerapp.py
@@ -32,6 +32,15 @@ options:
         description:
             - Resource location. If not set, location from the resource group will be used.
         type: str
+    kind:
+        description:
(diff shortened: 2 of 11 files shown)


plugins/module_utils/azure_rm_common.py
```diff
@@ -1340,6 +1342,15 @@ def containerinstance_client(self):
 
         return self._containerinstance_client
 
+    @property
+    def containerapps_client(self):
+        self.log('Getting Container Apps client')
+        if not self._containerapps_client:
+            self._containerapps_client = self.get_mgmt_svc_client(ContainerAppsAPIClient,
+                                                                  base_url=self._cloud_environment.endpoints.resource_manager,
+                                                                  api_version='2024-03-01')
```
Vieran: `api_version='2024-03-01'` is silently dropped here: `get_mgmt_svc_client` only forwards it when it's an explicit parameter of the client constructor (L948-L981):

But in `azure-mgmt-appcontainers==5.0.0` the signature is:
```
ContainerAppsAPIClient.__init__(self, credential, subscription_id, base_url=None, *, cloud_setting=None, **kwargs)

```

`api_version` is only reachable via `**kwargs`, so the value never reaches the client on L981. The effective version is the SDK default from `_configuration.py`: `kwargs.pop("api_version", "2026-01-01")`.

Please drop line 1351 (and update the PR description), or pin explicitly if an older API version is really required.

plugins/modules/azure_rm_containerapp.py
```diff
@@ -0,0 +1,1095 @@
...
+                        changed = True
+                        self.to_do = Actions.Update
+
+                # environment id / workload profile are stored in parameters using their
+                # ARM key names — compare against those directly.
+                for arm_key in ('managed_environment_id', 'workload_profile_name'):
+                    desired = self.parameters.get(arm_key)
+                    if desired is not None and desired != old_response.get(arm_key):
+                        changed = True
+                        self.to_do = Actions.Update
```
Vieran: This code set `to_do = Actions.Update` when `managed_environment_id` differs, but `_assign` (L989) deliberately keeps that key out of `update_parameters`, which means the update is a no-op and the module reports `changed=True` forever.

Since Microsoft.App can't move an app between environments, suggest dropping `managed_environment_id` from this loop and calling `self.module.warn()` instead (L950 is also a case-sensitive `!=` on a resource ID, which ARM casing alone can trip).

plugins/modules/azure_rm_containerapp_info.py
```diff
@@ -0,0 +1,189 @@
...
+            if self.show_secrets and self.name:
+                secrets = self._list_secrets()
+                # Register secret values with Ansible's redaction pool so they
+                # never appear in verbose (`-vvv`) or callback logs, even though
+                # they are returned in the module result for programmatic use.
+                for secret in secrets:
+                    value = secret.get('value')
+                    if value:
+                        self.module.no_log_values.add(value)
+                entry['secrets'] = secrets
```
Vieran, line 155: This injects a `secrets` key into each entry, but it's not documented anywhere in `RETURN` (L68-L75).

plugins/modules/azure_rm_containerapp.py
```diff
@@ -0,0 +1,1095 @@
...
+    if data.get('custom_domains') is not None:
+        data['custom_domains'] = [_clean(CustomDomain, c) for c in data['custom_domains']]
+    if data.get('ip_security_restrictions') is not None:
+        data['ip_security_restrictions'] = [_clean(IpSecurityRestrictionRule, r) for r in data['ip_security_restrictions']]
+    if data.get('cors_policy') is not None:
+        data['cors_policy'] = _clean(CorsPolicy, data['cors_policy'])
+    if data.get('sticky_sessions') is not None:
+        data['sticky_sessions'] = _clean(IngressStickySessions, data['sticky_sessions'])
+    if data.get('additional_port_mappings') is not None:
+        data['additional_port_mappings'] = [_clean(IngressPortMapping, p) for p in data['additional_port_mappings']]
```
Vieran, line 635: `sticky_sessions` and `additional_port_mappings` are handled here but aren't in `INGRESS_SPEC` (L750-L761), and since `ingress` is strictly validated via `options=INGRESS_SPEC` (L847), passing them just fails with `Unsupported parameters`, which means these branches are unreachable.

Same for `service_binds` at L670 (not in `TEMPLATE_KEYS`, L830) and `kind=params.get('kind')` at L699 (not in the arg spec). Either wire them into the arg spec + docs, or drop them along with the now-unused imports on L536-L538.

Vieran (changes requested):
Thanks for this! Mostly looks good. Some small comments below.

Vieran (approved):

PR #2354, opened 2026-08-26
New Module 866: Add azure_rm_loadbalancerbackendaddresspool and _info modules

### LINKED ISSUE #866: Allow Backend Address Members to be added to Pool in azure_rm_loadbalancer
I have not seen this mentioned so if It has I apologize. 

##### SUMMARY
I've found myself writing Azure CLI tasks to add members to pool, the pool has the ability in azure_rm_loadbalancer module to be created but I do not see any ability to add pool members. 

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure_rm_loadbalancer

##### ADDITIONAL INFORMATION
This feature would allow for a single module to be used with the ability to add/remove pool members within the same module. Currently if the azure_rm_loadbalancer module is used then another separate task with some Azure CLI needs to be written to allow the full deployment of a LB with pool members included. Basically adding in a function like the same code below. 

```yaml
- name: Create Load Balancer
  azure_rm_loadbalancer:
    resource_group: rg-sample
    name: lb-sample
    location: eastus
    sku: Standard
    frontend_ip_configurations:
      - name: fw-ip
        subnet: sn-subnet-name
        private_ip_allocation_method: Dynamic
    backend_address_pools:
      - name: be-pool-web
        members: # New feature
          - server: server1 # New feature
            ip: 10.0.0.10 # New feature
          - server: server2 # New feature
            ip: 10.0.0.11 # New feature
      - name: be-pool-sftp
        members: # New feature
          - server: server3 # New feature
            ip: 10.0.0.12 # New feature
          - server: server4 # New feature
            ip: 10.0.0.13 # New feature
    load_balancing_rules:
      - name: rule-web
        frontend_ip_configuration: fw-ip
        backend_address_pool: be-pool-web
        frontend_port: 443
        backend_port: 443
        probe: tcp-443
        idle_timeout: 30
        enable_floating_ip: false
        protocol: Tcp
        load_distribution: Default
      - name: rule-sftp
        frontend_ip_configuration: fw-ip
        backend_address_pool: be-pool-sftp
        frontend_port: 22
        backend_port: 22
        probe: tcp-22
        idle_timeout: 30
        enable_floating_ip: false
        protocol: Tcp
        load_distribution: Default
    probes:
      - name: tcp-443
        port: 443
        protocol: Tcp
      - name: tcp-22
        port: 22
        protocol: Tcp
    state: present

```

zunyangc:
##### SUMMARY
Fix #866.
Adds `azure_rm_loadbalancerbackendaddresspool` and `azure_rm_loadbalancerbackendaddresspool_info` modules to manage a load balancer's backend address pool (members, drain period, per-address admin state, VNet binding, tunnel interfaces) as a first-class resource. So users can populate a named pool stub created by `azure_rm_loadbalancer` without re-sending the full LB body.

##### ISSUE TYPE
- [x] New Module Pull Request
- [ ] Bug Fix Pull Request
- [ ] Test Fix Pull Request
- [ ] Migration Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
- plugins/modules/azure_rm_loadbalancerbackendaddresspool_info.py
- meta/runtime.yml
- tests/integration/targets/azure_rm_loadbalancerbackendaddresspool/

##### ADDITIONAL INFORMATION
- New CRUD module `azure_rm_loadbalancerbackendaddresspool` creates, updates, and deletes a single backend address pool on an existing Azure load balancer via `NetworkManagementClient.load_balancer_backend_address_pools`, avoiding a full-LB PUT on every membership change.
- Pool-level options: `virtual_network`, `sync_mode`, `drain_period_in_seconds`, `tunnel_interfaces`.
- Per-address options in `load_balancer_backend_addresses`: `name`, `ip_address`, `virtual_network`, `subnet`, `network_interface_ip_configuration`, `load_balancer_frontend_ip_configuration`, `admin_state`.
- New info module `azure_rm_loadbalancerbackendaddresspool_info` returns a single pool by name or lists all pools on a load balancer; returns an empty list when the named pool is not found.
- `meta/runtime.yml` registers both modules under `action_groups.azure`.
- Idempotence uses `AzureRMModuleBaseExt.default_compare` on a desired/current dict pair; per-address diffs are order-insensitive by `name`.
- Integration test target `azure_rm_loadbalancerbackendaddresspool` exercises create, idempotence, membership update (add/remove), `admin_state` and `drain_period_in_seconds` round-trip, `check_mode` non-mutation, info by-name, info list-all, info for non-existent pool, delete, and delete idempotence.

##### REFERENCE
- https://learn.microsoft.com/en-us/rest/api/load-balancer/load-balancer-backend-address-pools/create-or-update
- https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/network/azure-mgmt-network (`azure.mgmt.network.models.BackendAddressPool`, `LoadBalancerBackendAddress`, `GatewayLoadBalancerTunnelInterface`)

### DIFF
--- a/plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
+++ b/plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
@@ -1,0 +1,449 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# Copyright: (c) 2026, Zun Yang (@zunyangc).
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+---
+module: azure_rm_loadbalancerbackendaddresspool
+
+version_added: '4.0.0'
+
+short_description: Manage a backend address pool of an Azure load balancer
+
+description:
+    - Create, update or delete a single backend address pool on an existing Azure load balancer.
+    - The parent module M(azure.azcollection.azure_rm_loadbalancer) only declares empty pool stubs by name; this module owns all pool-level properties and members.
+
+options:
+    resource_group:
+        description:
+            - Name of the resource group that contains the load balancer.
+        required: true
+        type: str
+    load_balancer_name:
+        description:
+            - Name of the parent load balancer.
+        required: true
+        type: str
+    name:
+        description:
+            - Name of the backend address pool.
+        required: true
+        type: str
+    state:
+        description:
+            - Assert the state of the backend address pool.
+            - Use C(present) to create or update, or C(absent) to delete.
+        default: present
+        type: str
+        choices:
+            - absent
+            - present
+    virtual_network:
+        description:
+            - Resource ID of the virtual network the backend address pool is bound to.
+            - Required when I(sync_mode) is set.
+        type: str
+    sync_mode:
+        description:
+            - Backend address synchronous mode for the backend pool.
+            - Requires Standard SKU load balancer and I(virtual_network) to be set.
+        type: str
+        choices:
+            - Automatic
+            - Manual
+    drain_period_in_seconds:
+        description:
+            - Seconds the load balancer waits before sending RESET to client and backend address on connection close.
+        type: int
+    tunnel_interfaces:
+        description:
+            - Gateway load balancer tunnel interfaces for the backend address pool.
+        type: list
+        elements: dict
+        suboptions:
+            port:
+                description:
+                    - Port of the gateway load balancer tunnel interface.
+                type: int
+            identifier:
+                description:
+                    - Identifier of the gateway load balancer tunnel interface.
+                type: int
+            protocol:
+                description:
+                    - Protocol of the gateway load balancer tunnel interface.
+                type: str
+                choices:
+                    - None
+                    - Native
+                    - VXLAN
+            type:
+                description:
+                    - Traffic type of the gateway load balancer tunnel interface.
+                type: str
+                choices:
+                    - None
+                    - Internal
+                    - External
+    load_balancer_backend_addresses:
+        description:
+            - Backend addresses (IP-based or NIC-config-based) for the backend pool.
+            - Requires Standard SKU load balancer.
+        type: list
+        elements: dict
+        suboptions:
+            name:
+                description:
+                    - Name of the backend address, unique within the pool.
+                type: str
+                required: true
+            ip_address:
+                description:
+                    - IP address belonging to the referenced I(virtual_network).
+                type: str
+            virtual_network:
+                description:
+                    - Resource ID of the virtual network the backend address belongs to.
+                type: str
+            subnet:
+                description:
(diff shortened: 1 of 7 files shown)


plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
```diff
@@ -0,0 +1,451 @@
...
+            ) if self.virtual_network else None,
+            sync_mode=self.sync_mode,
+            drain_period_in_seconds=self.drain_period_in_seconds,
+            tunnel_interfaces=[self.network_models.GatewayLoadBalancerTunnelInterface(
+                port=ti.get('port'),
+                identifier=ti.get('identifier'),
+                protocol=ti.get('protocol'),
+                type=ti.get('type'),
+            ) for ti in self.tunnel_interfaces] if self.tunnel_interfaces else None,
+            load_balancer_backend_addresses=[self.network_models.LoadBalancerBackendAddress(
```
Vieran, line 396: This is a PUT, but `build_pool_param()` sends `null` for every option the user omitted. So setting only `drain_period_in_seconds` reports a one-field diff and then silently wipes the pool's entire membership. Unspecified fields need to be back-filled from the existing pool before the PUT.

plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
```diff
@@ -0,0 +1,451 @@
...
+        changed = False
+        response = existing_dict
+
+        if self.state == 'present':
+            if existing is None:
+                changed = True
+                if not self.check_mode:
+                    response = self.create_or_update()
+            else:
+                if not self.default_compare({}, desired, existing_dict, '', dict(compare=[])):
```
Vieran: Since backend addresses always have `name`, `default_compare` takes the `key = 'name'` branch, which never actually sorts, so the lists are compared positionally. Same members in a different order than Azure returns them means permanent `changed: true` plus a redundant PUT every run. Please sort by `name` on both sides before comparing.

plugins/modules/azure_rm_loadbalancerbackendaddresspool.py
```diff
@@ -0,0 +1,451 @@
...
+        default: present
+        type: str
+        choices:
+            - absent
+            - present
+    virtual_network:
+        description:
+            - Resource ID of the virtual network the backend address pool is bound to.
+            - Required when I(sync_mode) is set.
+        type: str
```
Vieran, line 53: The docs promise "Required when `sync_mode` is set" for `virtual_network`, but nothing enforces it, so `required_by=dict(sync_mode=['virtual_network'])` here would catch it locally instead of failing on the PUT.

Same for `backend_address_spec`: `ip_address`, `network_interface_ip_configuration` and `load_balancer_frontend_ip_configuration` are three mutually exclusive ways to address a member, and `tunnel_interfaces` needs all four suboptions together, so both could use `mutually_exclusive` / `required=True`.

Vieran (changes requested):
Several comments left as below.

Vieran (approved):

Issue #2357, opened 2026-09-02
azure_rm_azurefirewall: SDK refactor and API catch-up


zunyangc:
##### SUMMARY
`azure_rm_azurefirewall` has drifted significantly behind the current `Microsoft.Network/azureFirewalls` REST contract. It still calls the raw REST endpoint through `GenericRestClient` and hand-serialises `properties.*`, which is why every new API-version field has to be added by hand and why several long-standing feature gaps (DNS proxy, SNAT private ranges, IP groups on rules, Firewall Policy attachment, zones, autoscale, virtual-hub firewalls, etc.) never landed.

This issue tracks the work to bring the module up to the current API surface. It is a **superset umbrella** for smaller gap-fill issues; the DNS proxy piece was already handled in #304.

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
`azure.azcollection.azure_rm_azurefirewall`

##### ADDITIONAL INFORMATION
Reference contract used for the gap analysis below:
- REST: [Azure Firewalls - Create Or Update](https://learn.microsoft.com/en-us/rest/api/firewall/azure-firewalls/create-or-update) (Microsoft.Network, latest stable `2025-07-01`; module currently pins `api-version=2024-01-01`).
- SDK: `azure-mgmt-network` -> `NetworkManagementClient.azure_firewalls`.
- CLI cross-check (for the `additionalProperties`-map key strings): [`azure-cli-extensions/src/azure-firewall/azext_firewall/custom.py`](https://github.com/Azure/azure-cli-extensions/blob/main/src/azure-firewall/azext_firewall/custom.py) -> `AzureFirewallCreate.pre_operations`.

### 1. Architectural
- [ ] Switch the module from `GenericRestClient` + hand-crafted JSON to `azure.mgmt.network` SDK models (`NetworkManagementClient.azure_firewalls.begin_create_or_update`). This is the root cause of the drift and every field-level task below becomes cheap once this lands.
- [ ] Bump `api-version` from `2024-01-01` to the current stable (`2025-07-01`).
- [ ] Fix idempotence: current `default_compare({}, self.body, old_response, '', dict(compare=[]))` with an empty `compare=[]` treats every server-side default as a change on the next run. Adopt the `default_compare` pattern used by the sibling SDK-based modules.
- [ ] Return the full firewall envelope in `RETURN` instead of only `id`.

### 2. Top-level `properties.*` fields the module does NOT expose
| REST field | Suggested option | Why |
[...]

PR #2358, opened 2026-09-02
Migration 2357 & New feature 304: SDK refactor azure_rm_azurefirewall + DNS proxy support

### LINKED ISSUE #304: No DNS proxy settings avaulable for Azure firewall module
##### SUMMARY
It's not possible to define DNS proxy for Azure firewall via Ansible module

##### ISSUE TYPE
- Feature Idea

##### COMPONENT NAME
azure.azcollection.azure_rm_azurefirewall 

##### ADDITIONAL INFORMATION
To leverage FQDN on network rules we need DNS proxy enabled, when enabled the firewall will listen on port 53 and will forward DNS requests to the DNS server specified above. This will allow the firewall to translate that FQDN automatically.

zunyangc:
##### SUMMARY
Fix #304, #2357.
Both `azure_rm_azurefirewall` and `azure_rm_azurefirewall_info` move from the deprecated ARM `deploy_template` lifecycle to native `azure_firewalls` operations on `azure-mgmt-network`. The stateful module gains `dns_servers` and `dns_proxy_enabled` options wired through the SDK's `additional_properties` map, and `exec_module` now back-fills every unmanaged top-level field before PUT so partial updates cannot silently wipe firewall state.

##### ISSUE TYPE
- [x] New Feature Pull Request
- [ ] New Module Pull Request
- [ ] Bug Fix Pull Request
- [ ] Test Fix Pull Request
- [x] Migration Pull Request

##### COMPONENT NAME
- plugins/modules/azure_rm_azurefirewall.py
- plugins/modules/azure_rm_azurefirewall_info.py
- tests/integration/targets/azure_rm_azurefirewall/tasks/main.yml

##### ADDITIONAL INFORMATION
- Both modules migrated from `deploy_template` ARM lifecycle to `azure_firewalls.begin_create_or_update` / `.get` / `.list` / `.list_all` on `azure-mgmt-network` v2024-01-01.
- New options on `azure_rm_azurefirewall`: `dns_servers` (list of str) and `dns_proxy_enabled` (bool), persisted via `additional_properties` keys `Network.DNS.Servers` and `Network.DNS.EnableProxy` per the Azure Firewall REST spec.
- `exec_module` back-fills every top-level `AzureFirewall` SDK field the user did not supply (`application_rule_collections`, `nat_rule_collections`, `network_rule_collections`, `ip_configurations`, `sku`, `firewall_policy`, `threat_intel_mode`, `virtual_hub`, `zones`, `hub_ip_addresses`, `management_ip_configuration`, `autoscale_configuration`) from the existing firewall before PUT, so partial updates cannot clear collections or unmanaged state the user did not intend to touch.
- `additional_properties` is merged with the existing firewall so out-of-band keys (SNAT ranges, FTP flags, threat-intel allowlist, etc.) survive Ansible updates.
- Rule collections, their rules, and ip_configurations are pre-sorted by `name` before `default_compare` so Azure's return order no longer produces false-positive `changed=true`.
- `resolve_public_ip_id` accepts a bare name in the current resource group and constructs the full ARM ID via `azure.mgmt.core.tools.resource_id`; ARM IDs pass through unchanged via `is_valid_resource_id`.
- `resolve_subnet_id` validates the string form via `is_valid_resource_id` and fails fast with a clear error when given a bare non-ID name, matching the sibling `is_valid_resource_id` pattern in `azure_rm_appgateway`, `azure_rm_networkinterface`, and `azure_rm_subnet`.
- `RETURN` blocks on both modules now document `additional_properties`, `sku`, and `threat_intel_mode` to match `firewall.as_dict()`.
- 404-as-absent: `.get()` catches `ResourceNotFoundError` and returns `None` (stateful) / `[]` (info) so missing-firewall reads and delete-idempotence work correctly.
- Breaking: `azure_rm_azurefirewall_info` now returns `firewalls` as a list of dicts (matching the module's actual pre-refactor code behavior and the fixed RETURN doc). Playbooks that indexed `firewalls['someName']` must switch to `firewalls | selectattr('name', 'equalto', 'foo') | first`.
- Integration coverage extended: state envelope shape, info envelope, DNS proxy CRUD, drift preservation, reorder idempotence, partial-body back-fill, delete idempotence, missing-firewall info.

##### REFERENCE
- https://github.com/Azure/azure-sdk-for-python/tree/main/sdk/network/azure-mgmt-network/azure/mgmt/network/v2024_01_01
- https://github.com/Azure/azure-rest-api-specs/blob/main/specification/network/resource-manager/Microsoft.Network/stable/2024-01-01/azureFirewall.json

### DIFF
--- a/plugins/modules/azure_rm_azurefirewall.py
+++ b/plugins/modules/azure_rm_azurefirewall.py
@@ -357,6 +357,15 @@ EXAMPLES = '''
 '''
 
 RETURN = '''
+id:
+    description:
+        - The Azure Firewall resource ID. Preserved as a top-level field for
+          backward compatibility with pre-v4.x playbooks; the same value is
+          also available at I(state.id).
+    returned: success
+    type: str
+    sample: >-
+        /subscriptions/xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx/resourceGroups/myResourceGroup/providers/Microsoft.Network/azureFirewalls/myAzureFirewall
 state:
     description:
         - Current state of the Azure Firewall.
@@ -706,8 +715,6 @@ class AzureRMAzureFirewalls(AzureRMModuleBaseExt):
                 if update_tags:
                     changed = True
                     desired.tags = new_tags
-                elif self.tags is None:
-                    desired.tags = existing.tags
 
                 new_dict = desired.as_dict()
                 old_dict = existing.as_dict()
@@ -727,6 +734,8 @@ class AzureRMAzureFirewalls(AzureRMModuleBaseExt):
 
         self.results['changed'] = changed
         self.results['state'] = existing.as_dict() if existing else {}
+        # Preserve the `id` alongside the new `state` envelope
+        self.results['id'] = existing.id if existing else None
         return self.results
 
     def get_firewall(self):
--- a/plugins/modules/azure_rm_azurefirewall.py
+++ b/plugins/modules/azure_rm_azurefirewall.py
@@ -244,6 +244,16 @@ options:
                     - Name of the resource that is unique within a resource group.
                     - This name can be used to access the resource.
                 type: str
+    dns_servers:
+        description:
+            - List of custom DNS server IP addresses used by the firewall.
+        type: list
+        elements: str
+    dns_proxy_enabled:
+        description:
+            - Whether DNS proxy is enabled on the firewall.
+            - When C(true), the firewall listens on port 53 and forwards DNS queries to the addresses in I(dns_servers).
+        type: bool
     state:
         description:
             - Assert the state of the AzureFirewall.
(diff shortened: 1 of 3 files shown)


zunyangc:
Hi @Vieran, pipeline [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=344972&view=results) passed. Ready to merge, thanks!

plugins/modules/azure_rm_azurefirewall.py
```diff
@@ -557,340 +653,294 @@ def __init__(self):
...
+                              'management_ip_configuration', 'autoscale_configuration'):
+                    if getattr(desired, field, None) is None:
+                        setattr(desired, field, getattr(existing, field, None))
+
+                update_tags, new_tags = self.update_tags(existing.tags or {})
+                if update_tags:
+                    changed = True
+                    desired.tags = new_tags
+                elif self.tags is None:
+                    desired.tags = existing.tags
```
Vieran: With `append_tags: true`, unchanged subset tags can delete existing tags. Please always assign `new_tags` to `desired.tags`.

plugins/modules/azure_rm_azurefirewall.py
```diff
@@ -557,340 +653,294 @@ def __init__(self):
...
+        self.results['state'] = existing.as_dict() if existing else {}
         return self.results
 
-    def create_update_resource(self):
-        # self.log('Creating / Updating the AzureFirewall instance {0}'.format(self.))
+    def get_firewall(self):
+        try:
+            return self.network_client.azure_firewalls.get(self.resource_group, self.name)
+        except ResourceNotFoundError:
+            return None
```
Vieran, line 745: These SDK calls default to `2024-03-01`. Please explicitly use `2024-01-01` or document the upgrade.

plugins/modules/azure_rm_azurefirewall.py
```diff
@@ -557,340 +653,294 @@ def __init__(self):
...
-                response = self.get_resource()
-
+            if existing is not None:
+                changed = True
+                if not self.check_mode:
+                    self.delete_firewall()
+                    existing = None
+
+        self.results['changed'] = changed
+        self.results['state'] = existing.as_dict() if existing else {}
```
Vieran, line 736: Moving the resource ID from `result.id` to `result.state.id` may be a breaking change for existing playbooks. Please preserve the top-level `id`.

plugins/modules/azure_rm_azurefirewall_info.py
```diff
@@ -46,374 +51,161 @@
...
-                    '/resourceGroups' +
-                    '/{{ resource_group }}' +
-                    '/providers' +
-                    '/Microsoft.Network' +
-                    '/azureFirewalls' +
-                    '/{{ azure_firewall_name }}')
-        self.url = self.url.replace('{{ subscription_id }}', self.subscription_id)
-        self.url = self.url.replace('{{ resource_group }}', self.resource_group)
-        self.url = self.url.replace('{{ azure_firewall_name }}', self.name)
+        self.results['firewalls'] = [self.firewall_to_dict(item) for item in results]
```
Vieran: This changes both the named `result` from a `dict` to a `list` and nested fields from ARM-style to flattened snake_case. This is a broader breaking change. This should be preserved or deferred to a major release.

Vieran (changes requested):
This code change might contain breaking change. Those issues should be fixed before merge. Thanks!

plugins/modules/azure_rm_azurefirewall.py
zunyangc: Kept the sibling-parity pattern here for consistency with the other modules using the same shape ([`azure_rm_appgateway.py#L2390`](https://github.com/ansible-collections/azure/blob/dev/plugins/modules/azure_rm_appgateway.py#L2390) and family).

plugins/modules/azure_rm_azurefirewall.py
zunyangc: `self.network_client` is centrally pinned to `2024-01-01` at [`azure_rm_common.py#L1117-L1123`](https://github.com/ansible-collections/azure/blob/dev/plugins/module_utils/azure_rm_common.py#L1117-L1123), and every operation in this PR resolves against that pin.

plugins/modules/azure_rm_azurefirewall_info.py
Vieran: `_rule_to_arm` / `_ip_configuration_to_arm` rebuild the ARM shape by hand and silently drop fields the old pass-through returned, which includes `sourceIpGroups` (all three rule types), `translatedFqdn` (NAT), `destinationIpGroups` (network), `privateIPAddress` (ipConfigurations). And `privateIPAllocationMethod` is hardcoded to `'Dynamic'` even though the SDK model has no such attribute. Could you please fix this?

zunyangc: Updated, [test](https://dev.azure.com/azclitools/internal/_build/results?buildId=346613&view=results) passed. Thanks

Vieran (approved):

PR #2365, opened 2026-09-12
Fix canonical_facts and the capture() guards in the audit query

fistilis:
##### SUMMARY

Nothing in this repository executes the jq in `extensions/audit/event_query.yml`. It runs inside the Automation Platform controller ([`awx/main/tasks/host_indirect.py`](https://github.com/ansible/awx/blob/devel/awx/main/tasks/host_indirect.py)) and its output feeds subscription node counting, so a defect here is invisible from inside the collection. Four of them:

##### 1. `name` was in the dedup key, with an `"UNKNOWN"` fallback

`canonical_facts` is the **only** key the controller deduplicates on (`results[hashable_facts]`), and `facts` and `name` are taken from the first record seen. `canonical_facts: {name: ($data.name // "UNKNOWN"), id: ..., node_type: ...}` means the same resource counts twice whenever one module returns its name and another reaches it as a nested reference carrying only the id:

```
HEAD   one NIC, once with a name and once without -> 2 audit rows
  {"id": "/subscriptions/0000/resourceGroups/rg1/providers/Microsoft.Network/networkInterfaces/nic1", "name": "UNKNOWN", "node_type": "network"}
  {"id": "/subscriptions/0000/resourceGroups/rg1/providers/Microsoft.Network/networkInterfaces/nic1", "name": "nic1",    "node_type": "network"}

fixed  -> 1 audit row
  {"id": "/subscriptions/0000/resourceGroups/rg1/providers/Microsoft.Network/networkInterfaces/nic1", "node_type": "network"}
```

An ARM resource id is a complete identity on its own. `name` moves to `facts`, which is not hashed.

##### 2. `capture(...)?` yields the *empty stream* on no match, not null

```jq
(($data.id | capture("/providers/[Mm]icrosoft.(?<resourcetype>[^/]+)/")? | .resourcetype) | ascii_downcase) as $node_type |
```

An empty stream bound with `as` produces **no record at all**. So any resource whose id has no `/providers/Microsoft.X/` segment silently emitted nothing:

```
resourcegroup  (no output -- no audit record)
subscription   (no output -- no audit record)
```

Both captures now end in `// "unknown"`, and the filter that was happening by accident is written down explicitly. Resource groups and subscriptions stay excluded — they are containers, not nodes — but now *because the query says so*, not because `capture` returned an empty stream two lines earlier.

##### 3. The null guard ran after the thing it was guarding

```jq
($data.id | capture(...)? | ...) as $node_type |
($data.id | capture(...)? | ...) as $sub_node_type |
select($data.id != null) |          # <- too late
```

`$data.id` had already been passed to `capture()` twice. It now runs first, and checks the type as well: `capture()` raises on a non-string, and the controller catches jq errors at the **event** level — so one bad value discards every *other* module's audit records for that event, not just this one.

##### 4. The emitted `name` was the full ARM resource id

That is the string that appears in subscription reporting. It is now `($data.name // $data.id)`:

```
HEAD   display name: '/subscriptions/0000/resourceGroups/rg1/providers/Microsoft.Network/net…'
fixed  display name: 'nic1'
```

##### ALSO

Escapes the `.` in `Microsoft.` (it was an unescaped regex dot), and lowercases the id once up front instead of matching `[Mm]` and calling `ascii_downcase` on each capture result.

##### THE TEST

`tests/unit/audit/test_event_query_contract.py` asserts the controller's constraints statically:

- `name` present and cannot evaluate to null — `if name is None: continue`
- `canonical_facts` present and truthy — `if not data.get('canonical_facts'): continue`
- no field in `canonical_facts` can be null at any depth — `get_hashable_form()` accepts `int/float/str/bool/dict/list/tuple` and raises `UnhashableFacts` on anything else; the caller catches it and skips the record, logging once per job at `INFO` with the job still green
- `canonical_facts` holds identity only — this is the check that fails on `name`
- `facts.infra_type` / `infra_bucket` / `device_type` present and `lowercase_with_underscores`

It understands `//` chains, `select()` guards, `(… // "literal") as $var` bindings and `capture(…)?`, so it does not false-positive on the guarded fallbacks this file relies on. stdlib + pytest + PyYAML, no credentials, no Azure subscription, 0.05s.

##### NOTE

I did not touch `CHANGELOG.md` — the top section is the released `v4.0.0` and I did not want to invent a version heading. Happy to add an entry in whatever form you prefer.

Tracked internally as [AAP-92645](https://issues.redhat.com/browse/AAP-92645); part of a cross-collection audit of every collection that reports indirect nodes ([AAP-92620](https://issues.redhat.com/browse/AAP-92620)).

##### ISSUE TYPE
- Bugfix Pull Request

##### COMPONENT NAME
event_query

### DIFF
--- a/extensions/audit/event_query.yml
+++ b/extensions/audit/event_query.yml
@@ -1,4 +1,35 @@
 ---
+# Indirect node counting query for the azure.azcollection collection.
+#
+# This jq runs inside the Automation Platform controller
+# (awx/main/tasks/host_indirect.py) against each module's result, and its output
+# feeds subscription node counting. Nothing in this repository executes it, so
+# tests/unit/audit/test_event_query_contract.py asserts the contract the
+# controller imposes on the output.
+#
+# THREE THINGS THAT ARE EASY TO GET WRONG HERE
+# --------------------------------------------
+# 1. canonical_facts is the ONLY key the controller deduplicates on
+#    (`results[hashable_facts]`), and `facts` and `name` are taken from the
+#    first record seen. So it must hold identity and nothing else. A resource's
+#    ARM id is that identity on its own; anything else in canonical_facts can
+#    only split one resource into several rows. `name` used to be in here with
+#    an "UNKNOWN" fallback, so the same NIC reported once through
+#    azure_rm_networkinterface (name present) and once as a nested reference in
+#    another resource's result (name absent) produced two audit rows.
+#
+# 2. canonical_facts cannot contain null at any depth. get_hashable_form()
+#    accepts int/float/str/bool/dict/list/tuple and raises UnhashableFacts on
+#    anything else; the caller catches it and skips the record, logging once
+#    per job at INFO with the job still green. So every value below must be
+#    provably non-null, which is why the captures end in `// "unknown"`.
+#
+# 3. `capture(...)?` yields the EMPTY STREAM on no match, not null -- and an
+#    empty stream bound with `as` produces no record at all. That is why a
+#    resource group id, which has no /providers/Microsoft.X/ segment, silently
+#    emitted nothing. The filter is now explicit: only ARM resource ids are
+#    reported, and resource groups and subscriptions (which are containers,
+#    not nodes) are excluded on purpose rather than by accident.
 azure.azcollection.*:
   query: >-
     (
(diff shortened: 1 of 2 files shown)


p3ck (approved):
LGTM

zunyangc (approved):

zunyangc:
Hi @Vieran, ready for your review. Thanks!

extensions/audit/event_query.yml
```diff
@@ -178,26 +209,28 @@ azure.azcollection.*:
...
         (
-          ($data.id | capture("/providers/[Mm]icrosoft.[^/]+/(?<subtype>[^/]+)/")? | .subtype)
-            | ascii_downcase
+          ($arm_id | capture("/providers/microsoft\\.[^/]+/(?<subtype>[^/]+)/")? | .subtype)
+          // "unknown"
         ) as $sub_node_type |
-        select($data.id != null) |
         {
-          name: $data.id,
+          name: ($data.name // $data.id),
```
Vieran, line 224: Two different resources can have the same short name—for example, two VMs named `web` in different resource groups. AWX requires `(name, job)` to be unique, so a job containing both would fail to save its audit records. Could we keep `name` unique (for example, use the full ARM ID) and put the friendly name in `facts.name`?

extensions/audit/event_query.yml
```diff
@@ -178,26 +209,28 @@ azure.azcollection.*:
               else empty end
        end
        ) as $data |
+        select(($data.id | type) == "string") |
+        ($data.id | ascii_downcase) as $arm_id |
+        select($arm_id | test("/providers/microsoft\\.")) |
```
Vieran, line 214: This also matches provider-only IDs such as `/subscriptions/.../providers/Microsoft.Network`. They are not resource nodes, but the new `unknown` fallbacks cause them to be counted. Could we require a resource type and name after the provider?

Vieran (changes requested):
Mostly LGTM. Small suggestions left. Thanks!

PR #2366, opened 2026-09-14
Azure rm customerinsights instance

p3ck:
##### ISSUE TYPE
- New Module Pull Request

##### COMPONENT NAME
azure_powerplatform_customerinsights

##### ADDITIONAL INFORMATION

I'm having trouble testing this because my account doesn't give me access to power platform.  I've tried reaching out to my contact at MS but have not heard back..  I'm pushing this in the hopes someone else may have the needed permissions to test this properly.


p3ck:
> Hi @p3ck, thanks for this PR.
> 
> Quick question. I noticed Customer Insights doesn't seem to be an ARM resource (no rg, no Azure SDK backing it, and this PR calls the api directly). Just curious about the thinking here, is this meant to fit alongside the ARM-based modules?
> 
> Thanks!

It's not but seems to have a lot of overlap and I don't have a better place to put it.

For example here is the doc I have on how to setup permissions between azure and power platform: https://developer.ci.ai.dynamics.com/appregistration

I understand if you don't think it should be in this collection.

Vieran:
Thanks for opening this.

It looks like that this is outside the current scope of `azure.azcollection`. Customer Insights is a Dynamics 365 service, not an Azure resource, and we can’t currently test it. Maybe a dedicated Dynamics collection would be a better fit.

zunyangc:
Thanks @p3ck for this PR, closing this as not in scope.

